PR #12, opened 2022-05-15
CHIP-0006: DID External Identity Linking Specification

ddesiderio:
Created CHIP for DID External Identity Linking Specification


ddesiderio:
Due to lack of interest from community I'll be abandoning this CHIP

PR #13, opened 2022-05-18
CHIP-0003: Minimum Transaction Fee

danieljperry:
This is a draft proposal to add a minimum fee to Chia's full nodes.

### DIFF
--- a/CHIPs/chip-0003.md
+++ b/CHIPs/chip-0003.md
@@ -1,0 +1,174 @@
+CHIP Number   | 0003
+:-------------|:----
+Title         | Minimum Transaction Fee
+Description   | A method to add a minimum fee to Chia's full nodes
+Author        | [Dan Perry](https://github.com/danieljperry)
+Comments-URI  | < Creator must leave this blank. Editor will assign a URI.>
+Status        | Draft
+Category      | Process
+Sub-Category  | Environment
+Created       | 2022-05-11
+Requires      | None
+Replaces      | None
+Superseded-By | none
+
+## Abstract
+This CHIP will add a minimum, adjustable fee to the design of Chia's full node. The full nodes that choose to implement this feature will ignore any transactions which do not meet a minimum fee threshold.
+
+## Motivation
+Some background:
+
+When a user submits a transaction on Chia's blockchain, the transaction gets propagated across the network and added to the [mempool](https://docs.chia.net/docs/06mempool/mempool "Chia's mempool documentation") of each full node that receives it. When a farmer finds a winning proof for a transaction block, the farmer's full node chooses which transactions to add to the new block. Priority is given to the transactions with the highest fees (measured in mojos per [CLVM cost](https://chialisp.com/docs/ref/clvm#costs "Explanation of CLVM costs")).
+
+A few notes on the full node's current implementation:
+
+* There is no minimum fee to use Chia's blockchain, so farmers will automatically accept all incoming transactions, as long as their mempool isn't full
+* The mempool's maximum capacity is ten full blocks, or around `20 000` [vanilla transactions](https://chialisp.com/docs/standard_transaction "The standard transaction in Chia")
+* If a farmer's mempool is full, then that farmer's full node has two options for how to treat incoming transactions:
+  1. If the incoming transaction's fee is higher than that of the lowest priority transaction currently in the mempool, then the new transaction will be added to the mempool, prioritized by mojos per cost. This transaction will also be propagated to each of the node's peers. The lowest priority transaction(s) will be removed from the mempool
+  2. If the incoming transaction's fee is not higher than that of the lowest priority transaction currently in the mempool, then the new transaction will be ignored
+
+### Challenges
+This CHIP is an attempt to solve two limitations with the full node's current design:
+
+* **Challenge 1:** The fact that there is no minimum fee makes Chia's blockchain susceptible to dust storms, where a user spends many coins worth as little as one mojo (often referred to as "dust"). This is not considered an attack on Chia's network because the "dust stormer" is using the network within the bounds of its design. However, this _is_ an attack on individual users:
+  * Several users' wallets have become filled with thousands of dust coins. These coins have almost no value -- even a million mojos is worth significantly less than one US cent. Such wallets have long transaction histories, which slows syncing and makes record-keeping cumbersome
+  * Users who wish to send non-dust transactions often don't know how big of a fee to include. If they don't add a fee, their transaction might require an hour or more to be accepted. If they include a large fee, their transaction will be accepted quickly, but the fee might have been orders of magnitude higher than it needed to be
+
+* **Challenge 2:** In the medium term, as Chia's network grows in popularity, transaction fee pressure is expected to mount. This will naturally result in a minimum fee to use the network. If this CHIP has not been implemented by that point, then zero-fee transactions could get stuck in the mempool indefinitely
+
+### Solution
+This CHIP will solve these problems by adding an adjustable fee to the full node's default config. This fee's default value will be `181 818` mojos per million cost. This will be the minimum amount required for the full node to route transactions to the mempool.
+
+The resulting fee will be around 1 million mojos for a one-input, one-output transaction. This specific amount was chosen to make it costly to spam the network with dust transactions, while still being cheap enough to support pools and faucets. It will also prepare the network for a future with significantly more non-dust usage.
+
+### Benefits
+* Dust stormers will need to pay a fee to use the network. Transactions of a single mojo will still be allowed, but it will now cost around 1 million mojos for the privilege
+* Farmers will receive increased revenue due to a minimum fee being added to transactions
+
+### Drawbacks and Rebuttals
+1. **New Users and Faucets**
+    * **Drawback**: One mojo is required to create a singleton for pooling. New users can acquire _some_ mojos from faucets, but they might not receive a sufficient amount to pay a transaction fee for pool participation
+    * **Rebuttal**: The minimum fee is sufficiently small to allow faucets to continue operating, even after they increase their payouts to provide a fee for new users. As explained in the [Rationale](#rationale) section, a faucet funded with 1 XCH (currently $39) will be able to make `250 000` payouts to customers. These payouts include the fee required to create a singleton
+
+2. **Pool Operators**
+    * **Drawback**: Large pools typically send frequent payments to their farmers. If they need to add a fee to each of these payments, they might not be able to justify making them as frequently
+    * **Rebuttal**: We intentionally set the minimum fee to be quite low. It will cost around 2 million mojos per payment, which equates to $0.000078 at the current exchange rate of $39 per XCH. This should be sufficiently cheap for pools to make regular payments to their customers, even those with small farms
+
+3. **Wallet Developers**
+    * **Drawback**: Wallet developers will be recommended to update their implementations to take the new minimum fees into account. If this CHIP is accepted and wallets choose not to enforce minimum fees, then users' transactions might not make it into the mempool, leading to end user frustration
+    * **Rebuttal**: Even if this CHIP were not implemented, a fee market is expected to emerge on Chia's blockchain in the medium term. Wallets will need to add logic to calculate a minimum fee regardless of the status of this CHIP
+
+4. **Farmers**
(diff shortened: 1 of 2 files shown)


CHIPs/chip-0003.md
```diff
@@ -0,0 +1,174 @@
...
+The first alternative proposal we considered was to add an adjustable fee, with a default value of `0`, to the full node's default config. Not only did we accept this proposal, but for several reasons, we decided to add it _prior_ to implementing this CHIP:
+* This change is non-controversial because of the default fee of `0`. If users do not change this setting, their full node's functionality will not be affected
+* This change will allow farmers to control the minimum transaction fee they are willing to accept for blocks that they farm
+* This change will be a stepping stone toward developing a fee market for Chia's ecosystem
+
+We also considered -- and rejected -- several alternative methods to mitigate the problems laid out at the beginning of this CHIP:
+
+1. Do nothing. This would maintain the status quo and avoid the aforementioned drawbacks, but we feel the benefits of this CHIP still make it worth implementing
+
+2. Censor the dust stormer. This proposal was a non-starter. Censorship is anathema to the ethos of Chia's blockchain. Additionally, it would break the Sybil-resistance guarantees of Chia's protocols
```
hoffmang9, line 111: I think we have to add that we couldn't do it even if we wanted to as it is technically infeasible.

hoffmang9 (approved):
lgtm

CHIPs/chip-0003.md
```diff
@@ -0,0 +1,174 @@
...
+
+### Challenges
+This CHIP is an attempt to solve two limitations with the full node's current design:
+
+* **Challenge 1:** The fact that there is no minimum fee makes Chia's blockchain susceptible to dust storms, where a user spends many coins worth as little as one mojo (often referred to as "dust"). This is not considered an attack on Chia's network because the "dust stormer" is using the network within the bounds of its design. However, this _is_ an attack on individual users:
+  * Several users' wallets have become filled with thousands of dust coins. These coins have almost no value -- even a million mojos is worth significantly less than one US cent. Such wallets have long transaction histories, which slows syncing and makes record-keeping cumbersome
+  * Users who wish to send non-dust transactions often don't know how big of a fee to include. If they don't add a fee, their transaction might require an hour or more to be accepted. If they include a large fee, their transaction will be accepted quickly, but the fee might have been orders of magnitude higher than it needed to be
```
mariano54, line 36: this is not really solved by adding a min fee. if anything it makes it more confusing for the user

CHIPs/chip-0003.md
```diff
@@ -0,0 +1,174 @@
...
+
+* **Challenge 2:** In the medium term, as Chia's network grows in popularity, transaction fee pressure is expected to mount. This will naturally result in a minimum fee to use the network. If this CHIP has not been implemented by that point, then zero-fee transactions could get stuck in the mempool indefinitely
```
mariano54, line 38: transactions don't get stuck in the mempool, they either go in the block, or they get kicked from the mempool.
Also, this is not solved by adding a min fee, since wallets might still try to send low fee txs.
This is solved by adding a "replace by fee" feature and a "fee estimation" feature.

CHIPs/chip-0003.md
```diff
@@ -0,0 +1,174 @@
...
+
+### Drawbacks and Rebuttals
+1. **New Users and Faucets**
+    * **Drawback**: One mojo is required to create a singleton for pooling. New users can acquire _some_ mojos from faucets, but they might not receive a sufficient amount to pay a transaction fee for pool participation
+    * **Rebuttal**: The minimum fee is sufficiently small to allow faucets to continue operating, even after they increase their payouts to provide a fee for new users. As explained in the [Rationale](#rationale) section, a faucet funded with 1 XCH (currently $39) will be able to make `250 000` payouts to customers. These payouts include the fee required to create a singleton
+
+2. **Pool Operators**
+    * **Drawback**: Large pools typically send frequent payments to their farmers. If they need to add a fee to each of these payments, they might not be able to justify making them as frequently
+    * **Rebuttal**: We intentionally set the minimum fee to be quite low. It will cost around 2 million mojos per payment, which equates to $0.000078 at the current exchange rate of $39 per XCH. This should be sufficiently cheap for pools to make regular payments to their customers, even those with small farms
```
mariano54, line 56: This seems too low. It's about $150 per day to spam:
0.000078 * 2000 tx/block * 1000 txblocks 
I would increase this maybe 10x

CHIPs/chip-0003.md
```diff
@@ -0,0 +1,174 @@
...
+
+3. **Wallet Developers**
+    * **Drawback**: Wallet developers will be recommended to update their implementations to take the new minimum fees into account. If this CHIP is accepted and wallets choose not to enforce minimum fees, then users' transactions might not make it into the mempool, leading to end user frustration
+    * **Rebuttal**: Even if this CHIP were not implemented, a fee market is expected to emerge on Chia's blockchain in the medium term. Wallets will need to add logic to calculate a minimum fee regardless of the status of this CHIP
```
mariano54, line 60: Also, the user will immediately see that the TX was rejected

CHIPs/chip-0003.md
```diff
@@ -0,0 +1,174 @@
...
+* Mojos required to dust the network per day = `3 322 000` tx * `1 000 000` mojos fee per tx = `3 322 000 000 000`
+* Mojos per XCH = 1 trillion (`1 000 000 000 000`)
+* XCH required to dust the network per day = `3.322` ($129.56 at the current exchange rate of $39 per XCH). This assumes the duster already has a sufficient supply of dust coins, so there would be no need to break apart any large coins
+
+We feel that this would be a sufficiently high price to discourage dust. The minimum fee, in mojos per million cost, is then calculated as follows:
+* Maximum CLVM cost per block = `11 000 000 000`
+* Typical CLVM cost per 1-input, 1-output vanilla transaction = `11 000 000 000 / 2000` = `5 500 000`
+* Mojos per million cost required to achieve a `1 000 000` mojo minimum fee = `(1 000 000 / 5 500 000) * 1 000 000` = `181 818`
+
+We can also estimate the fees required for pools to pay their customers. The transactions would likely require two outputs (one coin for the receiver and one for "change"), so the CLVM cost would be higher than what was calculated above. We can use a conservative estimate of 1000 transactions per block.
```
mariano54, line 144: pool payous are batched so they contain only 1 ouput per user

danieljperry:
@mariano54 I have updated the CHIP, taking all of your suggestions into account.

CHIPs/chip-0003.md
```diff
@@ -0,0 +1,167 @@
...
+
+The resulting fee will be around 10 million mojos for a one-input, one-output transaction. This specific amount was chosen to make it costly to spam the network with dust transactions, while still being cheap enough to support pools and faucets. It will also prepare the network for a future with significantly more non-dust usage.
+
+### Benefits
+* Dust stormers will need to pay a fee to use the network. Transactions of a single mojo will still be allowed, but it will now cost around 10 million mojos for the privilege
+* Farmers will receive increased revenue due to a minimum fee being added to transactions
+
+### Drawbacks and Rebuttals
+1. **New Users and Faucets**
+    * **Drawback**: One mojo is required to create a singleton for pooling. New users can acquire _some_ mojos from faucets, but they might not receive a sufficient amount to pay a transaction fee for pool participation
```
Hadamcik, line 47: It would be great if this minimum fee change also supported X zero-fee transactions per sender address every Y blocks and whitelisting sender address.

- X per Y would allow people to set up/change DIDs, plotnfts, etc. for free and still prevent dust storms
- whitelisting would allow farmers to support specific senders (like faucets) in their activity without paying a fee

lsp0110psl:
When the price of XCH:USDS begins to vary significantly, then adjusting fee thresholds will become an issue.  The VWAP (Volume Weighted Average Price) of XCH/USDS over a recent time period (example: 24 hours) could be one coefficient used in determining such a minimum fee.  Example PsuedoCode:

XCHperUSDest = VWAP(tickerNum=XCH, tickerDenom=USDS, hrs=24)
noiseFloorEstimateMinCost = XCHperUSDest * FEE_PERCENTAGE

An alternative starting point, could be an UI solution for the user, by displaying: median times to execute transactions as a function of fee added during recent time frames (maybe displaying a graph and explaining this calculation).  When Chia Network usage increases, this will vary more than it does presently.

lsp0110psl:
I removed my initial VWAP (Volume Weighted Average Price) suggestion for adjusting minimum fee, because now I see complications in determining fair prices:

1[Fake Pricing]: There is no best-price-first-execution rule (direct offers and separate exchanges circumvent this rule).

2[Fake Volume]: Since the duster can amplify volume by sending excess trades back and forth between controlled wallets (direct offers), this extra volume would distort VWAP.

If it becomes possible to use data from a Decentralized Exchange (implicitly using a best-price-first-execution rule), then fair pricing can be statistically determined (by VWAP or other methods) and used to adjust the minimum fee.

I am not yet a code contributor, so I apologize if this is beyond scope or if you are already working internally on a better solution.

danieljperry:
> I removed my initial VWAP (Volume Weighted Average Price) suggestion for adjusting minimum fee, because now I see complications in determining fair prices:
> 
> 1[Fake Pricing]: There is no best-price-first-execution rule (direct offers and separate exchanges circumvent this rule).
> 
> 2[Fake Volume]: Since the duster can amplify volume by sending excess trades back and forth between controlled wallets (direct offers), this extra volume would distort VWAP.
> 
> If it becomes possible to use data from a Decentralized Exchange (implicitly using a best-price-first-execution rule), then fair pricing can be statistically determined (by VWAP or other methods) and used to adjust the minimum fee.
> 
> The Best Execution Rule is used in the conventional markets. It is reasonable to consider whether a similar rule will apply to DeFi crypto in the future.
> 
> https://www.investopedia.com/terms/b/bestexecution.asp
> 
> https://en.wikipedia.org/wiki/Best_execution

Thanks for adding these thoughts. This CHIP was not popular and likely won't be implemented. But I'm not quite ready to withdraw it just yet.

hoffmang9 (approved):
lgtm

KryptomineCH:
I disagree with this proposal to implement a minimum fee on Chia's blockchain because it goes against the principles of a free market and is an unnecessary intervention into the current functioning system. The current design with variable fees is a self-balancing system, and as adoption increases, the issue of dust storms will naturally resolve itself through higher fees. Additionally, implementing a minimum fee artificially limits the capabilities of the blockchain and may discourage legitimate usage and negatively impact the attractiveness of the blockchain to users.

Furthermore, this minimum fee could disproportionately affect small transactions and micropayments, making it less accessible for individuals and smaller businesses.

lsp0110psl:
> I disagree with this proposal to implement a minimum fee on Chia's blockchain because it goes against the principles of a free market and is an unnecessary intervention into the current functioning system. The current design with variable fees is a self-balancing system, and as adoption increases, the issue of dust storms will naturally resolve itself through higher fees. Additionally, implementing a minimum fee artificially limits the capabilities of the blockchain and may discourage legitimate usage and negatively impact the attractiveness of the blockchain to users.
> 
> Furthermore, this minimum fee could disproportionately affect small transactions and micropayments, making it less accessible for individuals, smaller businesses and startup projects.

~~~

Even if the proposal is withdrawn, the discussion has been useful.  I see commits on Github making improvements to the mempool optional fee estimator, which (as it is optional,) does not impede or add friction to the market.

I did not commit to being for or against this proposal; I only suggested ways to make it more fair should a minimum fee be enforced.

Having good data for statistically-determined recommended minimums should be sufficient IMHO.  I am currently a non-contributing observer.  Thank you for considering my point of view.

To restate a further refinement: at some point it would be useful to get data-feeds from DeFi Exchanges that are trusted to execute at best-price-first (XCH <-> Stable Coins), since it will be a common use case for many to convert to a Stable Coin.  That is a separate, but related use case, to optional mempool fee calculations.  Why is it related?  Because the user may be accustomed to thinking in terms of Stable Coins, not mojos; and should the Stable Coin conversion calculation be manipulated by a duster, this could potentially cause large accidental fees/"donations" through the user interface.  Even if Chia Inc clamps the maximum fee in the GUI to prevent this, eventually there will be widely-used 3rd-party GUIs, so again this problem resurfaces.

I see Chia Inc has partnered with Stably and Dexie to do XCH/USDS Transactions, so perhaps there is a future data-feed.  I have zero inside information on Chia Inc's future plans; only what I see on Github and Chia.net.  [https://dexie.space/get-usds](url)

Clean data = Happy users.

☮️

danieljperry:
Given that we now have both a dust filter and a working fee estimator (as of 1.7), and given that this proposal did not achieve consensus from the community, I am going to withdraw it from consideration.

PR #18, opened 2022-06-06
CHIP-0004: DID1 Standard

danieljperry:


ddesiderio (commented):
Thanks for getting this CHIP started to define the standard for DID1. I've added a few preliminary discussion points related to the spec for your review. Related to my previous CHIP request I have also asked a few questions around metadata and respective schema which would be great to solve before DID1 release.

Looking forward to seeing some more detail on the RPCs once this gets fleshed out a bit further 😄

CHIPs/chip-0004.md
```diff
@@ -0,0 +1,492 @@
...
+ 
+* **BLS signatures:** Chia uses [BLS signatures](https://docs.chia.net/docs/09keys/keys-and-signatures/), which have built-in capability for M of N multisig. This will allow Chia DIDs to be shared between multiple parties, and only allow transactions to occur if a threshold of signatures has been reached. BLS signatures also enable recovery of a lost key with M of N trusted signatures
+ 
+* **Offers:** Requests for credentials could be incorporated into [offers](https://chialisp.com/docs/puzzles/offers/), which are partially signed transactions with no central intermediary required for completion
+
+### Design decisions
+* Each DID that follows the DID1 standard is required to be implemented as a singleton
+* Because DIDs are singletons, a spend of at least 1 mojo is required to register a DID on-chain
+* Each DID must contain the owner's public key
```
Hadamcik, line 100: I think that multiple keys should be supported:

A person can have multiple keys, which can have a different level of security. It would be great if DID could recognize differences in this security (for example, key A is on my keychain, key B is in my home behind locked doors, key C is in a safe and requires a key, code, and biometrics,...).

- A lower-level key can be revoked using a higher-level key (or using a combination of some keys to get a higher security level). Keys can also be added
- DID signature weight should be determined by key security level (some actions may require a higher authentication level)
- DID signature should be revokable based on the key security level

Of course, for this to be possible, key security levels have to be universally defined.

ddesiderio: I think for the time being the plan for multi keys is to use the recovery mechanism- so you can recover a DID to another wallet/key if you have M of N signatures of known recovery keys. This doesn't give all the functionality you've described, but it does at least allow for some of the revoking behavior you're after.

CHIPs/chip-0004.md
```diff
@@ -0,0 +1,492 @@
...
+* **my_amount_or_inner_solution:**
+  * In mode 0, this is an amount. Recover this coin and assert its amount
+  * In mode 1, this is the solution to the inner p2 puzzle
+* **new_inner_puzhash:** In recovery mode, this will be the new wallet DID puzzle hash
+* **parent_innerpuzhash_amounts_for_recovery_ids:** A list of DIDs, each of which contains a complete list of arguments, to be used for recovery
+* **pubkey:** The new public key used for a recovery
+* **recovery_list_reveal:** This is the list of DIDs that had previously been approved for recovery. The hash of this list must equal RECOVERY_DID_LIST_HASH
+* **my_id:** This coin's coin_id
+
+### RPC Calls
```
mariano54, line 184: It seems like these RPCs are just what our wallet exposes, but they don't necessarily have to be implemented by other wallet or DID developers. If I make a DID wallet without these RPCs, it it now not compliant with the spec?

CHIPs/chip-0004.md
```diff
@@ -0,0 +1,492 @@
...
+ 
+* **Chialisp:** [Chialisp](https://chialisp.com/) is Chia's Turing-complete on-chain programming language. Like all Lisp variants, Chialisp can be represented as a binary tree. This design allows for an arbitrary amount of data to be used with each program, contained in a single tree hash. A DID's owner can therefore choose whatever information they want to be included in their DID. Chialisp also enables strong composability and interoperability between coins, creating the potential for nearly limitless customizable features
+ 
+* **BLS signatures:** Chia uses [BLS signatures](https://docs.chia.net/docs/09keys/keys-and-signatures/), which have built-in capability for M of N multisig. This will allow Chia DIDs to be shared between multiple parties, and only allow transactions to occur if a threshold of signatures has been reached. BLS signatures also enable recovery of a lost key with M of N trusted signatures
```
ddesiderio, line 93: Just to be clear about the recovery mechanism (as I understand it, and based on current state of the reference implementation) - it looks as though the purpose of this is _only_ to recovery DIDs from wallets with lost keys. It does not provide a sufficient clawback mechanism in the event that keys become compromised. This is because it looks as though there is no time limit to the recovery state, and the owner of the DID is able to update/remove recovery ids at any point. This means if a DID wallet has keys compromised a bad actor could just remove the recovery ids and prevent the original owner from recovering the wallet.

CHIPs/chip-0004.md
```diff
@@ -0,0 +1,492 @@
...
+* They can have unlimited scope
+  * Chia DIDs can be used by any marketplace that implements the open-source standard laid out in this document
+* They can be interoperable
+  * Different companies or entities could accept Chia DIDs, even if they were competitors
+
+Use cases for Chia DIDs include, but are not limited to:
+* Cryptographically proving identity online
+* Enabling marketplaces that are not walled gardens
+* Enabling self-custody of digital and physical assets
+* Enabling the designation of trusted entities for M-of-N wallet recovery in cases where the owner loses their original credentials
```
ddesiderio, line 57: This should be clear to call out this is only attempting to solve the problem of owner losing keys, but not keys being compromised. As the current proposal doesn't seem to provide security in that scenario (see my other comment)

CHIPs/chip-0004.md
```diff
@@ -0,0 +1,492 @@
...
+  * A DID's owner must be able to set and trigger the recovery process for their DID
+  * A DID's owner must be able to recover their DID on a different computer than the one on which it was created
+
+* Upgradeability
+  * DIDs using the DID1 specification must be upgradeable to include features of future DID specifications
+
+* Marketplace custody
+  * Marketplaces may custody DIDs on their owners' behalf
+  * Marketplaces must obtain permission from a DID's owner before custody is transferred
+  * This consent may not be withdrawn once given. The owner must trust the marketplace to return the DID upon request
```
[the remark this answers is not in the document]
danieljperry: The DID will need to be transferred, at least in this iteration. Clawback might be an idea for a future DID version, though.

CHIPs/chip-0004.md
```diff
@@ -0,0 +1,492 @@
...
+  * A DID owner may create more than one DID using the same public/private key pair
+  * A DID shall not have exclusive control over a public/private key pair
+  * A DID shall not own another DID
+  * DIDs may own multiple NFTs
+  * The DID hash may be shared as a means to verify identity
+  * The DID hash must be made discoverable by blockchain explorers
+  * DIDs must be able to discover the assets they own
+  * DID owners may provide their DIDs to issuers to be signed as Verifiable Credentials (VCs), thus allowing VCs to be issued to those DIDs
+  * DID owners must be able to view and copy their DID ID
+  * DID owners must be able to view the metadata associated with their DIDs
```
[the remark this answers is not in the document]
danieljperry: added

CHIPs/chip-0004.md
danieljperry: This is a good idea to propose for a future DID version.

CHIPs/chip-0004.md
danieljperry: Correct, this is primarily a mechanism to protect against loss of DID. _Theft_ of DID is another issue that we could potentially address with clawbacks in a future version.

CHIPs/chip-0004.md
danieljperry: Agreed and noted.

hoffmang9 (approved):
lgtm

CHIPs/chip-0004.md
```diff
@@ -0,0 +1,492 @@
...
+  * ((SINGLETON_MOD_HASH, (LAUNCHER_ID, LAUNCHER_PUZZLE_HASH))), where:
+    * **SINGLETON_MOD_HASH** is the tree hash of [singleton_top_layer_v1_1.clvm](https://github.com/Chia-Network/chia-blockchain/blob/main_dids/chia/wallet/puzzles/singleton_top_layer_v1_1.clvm) without any curried arguments
+    * **LAUNCHER_ID** is the coin_id of the singleton used with this DID
+    * **LAUNCHER_PUZZLE_HASH** is the tree hash of [singleton_launcher.clvm](https://github.com/Chia-Network/chia-blockchain/blob/main_dids/chia/wallet/puzzles/singleton_launcher.clvm), which creates the singleton coin and announcement
+  * **METADATA:** Customizable metadata, e.g. Know Your Customer (KYC) info
+
+Additionally, each DID's inner puzzle must include the following solution upon initiating a message creation, recovery or self-destruct action:
+* **mode:** The type of spend. Choose from
+  * 0 = Recovery
+  * 1 = Run INNER_PUZZLE with p2_solution
```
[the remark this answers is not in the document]
trepca: @ddesiderio sorry for late reply, but the idea here is not to have many modes, but to that inner puzzle can update itself to whatever it wants the inner puzzle to be. This makes DID freely to implement any puzzle / functionality it wants.

CHIPs/chip-0004.md
```diff
@@ -0,0 +1,492 @@
...
+* **RECOVERY_DID_LIST_HASH:** The hash of the list of DIDs that can be used to recover this DID by sending this DID a message. A hash is stored instead of the full list in order to avoid revealing the whole list every time the singleton is spent
+* **NUM_VERIFICATIONS_REQUIRED:** The number of DIDs from the above list that are required for recovery
+* **SINGLETON_STRUCT:** Defines the singleton used with this DID. Contains the following structure
+  * ((SINGLETON_MOD_HASH, (LAUNCHER_ID, LAUNCHER_PUZZLE_HASH))), where:
+    * **SINGLETON_MOD_HASH** is the tree hash of [singleton_top_layer_v1_1.clvm](https://github.com/Chia-Network/chia-blockchain/blob/main_dids/chia/wallet/puzzles/singleton_top_layer_v1_1.clvm) without any curried arguments
+    * **LAUNCHER_ID** is the coin_id of the singleton used with this DID
+    * **LAUNCHER_PUZZLE_HASH** is the tree hash of [singleton_launcher.clvm](https://github.com/Chia-Network/chia-blockchain/blob/main_dids/chia/wallet/puzzles/singleton_launcher.clvm), which creates the singleton coin and announcement
+  * **METADATA:** Customizable metadata, e.g. Know Your Customer (KYC) info
```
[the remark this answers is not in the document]
trepca: No specific format, but wallet RPC takes dictionary of (str, str).

CHIPs/chip-0004.md
```diff
@@ -0,0 +1,745 @@
...
+* **RECOVERY_DID_LIST_HASH:** The hash of the list of DIDs that can be used to recover this DID by sending this DID a message. A hash is stored instead of the full list in order to avoid revealing the whole list every time the singleton is spent
+* **NUM_VERIFICATIONS_REQUIRED:** The number of DIDs from the above list that are required for recovery
+* **SINGLETON_STRUCT:** Defines the singleton used with this DID. Contains the following structure
+  * ((SINGLETON_MOD_HASH, (LAUNCHER_ID, LAUNCHER_PUZZLE_HASH))), where:
+    * **SINGLETON_MOD_HASH** is the tree hash of [singleton_top_layer_v1_1.clvm](https://github.com/Chia-Network/chia-blockchain/blob/main/chia/wallet/puzzles/singleton_top_layer_v1_1.clvm) without any curried arguments
+    * **LAUNCHER_ID** is the coin_id of the singleton used with this DID
+    * **LAUNCHER_PUZZLE_HASH** is the tree hash of [singleton_launcher.clvm](https://github.com/Chia-Network/chia-blockchain/blob/main/chia/wallet/puzzles/singleton_launcher.clvm), which creates the singleton coin and announcement
+* **METADATA:** Customizable metadata, e.g. Know Your Customer (KYC) info
```
trepca, line 165: we could say this is a list of `(str, str)` as we currently support in in wallet RPC

danieljperry:
This CHIP is now Final. Further changes (other than errata) are no longer allowed.

hoffmang9 (approved):
lgtm

CHIPs/chip-0004.md
```diff
@@ -1,14 +1,14 @@
-CHIP Number   | < Creator must leave this blank. Editor will assign a number.>
```
VINELLO80: 004

PR #24, opened 2022-06-15
Auctions - Initial commit

joshpainter:
This CHIP details a zero-counterparty-risk protocol for auctions of digital assets running on the Chia blockchain. Ideally any auction protocol will be interoperable with all existing and future assets that may be created on the Chia blockchain, and therefore it is worth discussing and designing a solid foundation with full community involvement.

This first draft details the problems and risks with the current-day auction process. The next update will add the first draft for the Conceptual Design and Specification sections - the fun part!

Special thanks to @danieljperry and @fizpawiz for their help so far!

### DIFF
--- a/CHIPs/chip-joshpainter-auctions.md
+++ b/CHIPs/chip-joshpainter-auctions.md
@@ -1,0 +1,86 @@
+CHIP Number   | < Creator must leave this blank. Editor will assign a number.>
+:-------------|:----
+Title         | Auctions
+Description   | A method for holding decentralized, zero-risk auctions for digital assets on the Chia blockchain
+Author        | [Josh Painter](https://github.com/joshpainter) ([@endertown](https://twitter.com/endertown) on Keybase and Twitter)
+Comments-URI  | < Creator must leave this blank. Editor will assign a URI.>
+Status        | < Creator must leave this blank. Editor will assign a status.>
+Category      | Informational
+Sub-Category  | Guideline
+Created       | 2022-05-22
+Requires      | none
+Replaces      | none
+Superseded-By | none
+
+## Abstract
+Auction is one of the oldest methods of negotiating the exchange of goods and commodities between buyers and sellers. As early as 500 BC, Babylonians held annual auctions for attractive maidens who had reached marrying age. Ancient Greeks and Romans would auction off their spoils of battle to gain treasure for the war effort. Romans also used auctions to liquidate assets and pay off debt. Surprisingly, the entire Roman Empire was put up for auction on March 28th, 193 AD by the Praetorian Guard and purchased by Didius Julianus! Although it couldn't be built in a day, apparently Rome could at least be *purchased* within that timeframe. More recently, the Internet has enabled a massive new market for online auctions. Starting with eBay in 1995 and growing every year, online auctions continue to be an important method for trading assets to this day.
+
+However, even with all this history, auctions still have unsolved risks for the buyer, the seller, and especially the auction house holding the auction itself. The Chia blockchain, along with the coinset model and Chialisp, brings about new and unique opportunities to improve on this millenia-old practice and eliminate risk for all parties. This document will explore the counterparty risks and issues with modern-day auction. It will then propose high-level solutions to these problems using puzzles created in Chialisp and executed directly on the Chia blockchain. Finally, it will recommend detailed specifications that can be used to build these Chialisp puzzles to enable these solutions.
+
+## Motivation
+Modern-day auctions continue to be a valuable tool for finding market value of new assets, quickly liquidating old assets, and everything in between. However, multiple risks still exist for the bidders, the sellers, and especially the auction house responsible for promoting and organizing the auction itself.
+
+  * Seller risks
+    * High bidder may not complete transaction after auction ends. This forces seller to find another buyer who will usually not be willing to pay the same price. The best case is probably offering the auctioned item to the second-highest bidder at their second highest bid, but usually the second-highest bidder will recognize their negotiating power and drive the price even lower. Current solutions to this problem involve legal contracts and bidder deposits, both of which can add significant friction to the auction event itself. 
+    * High bidders might hold back bids until the last minute, hoping to “snipe” other users. Ideally users would be incentivized to make lots of small bids over the entire auction event instead of a few large bids in a short time near the end. This promotes more engagement with the auction itself and might result in a higher price for the seller.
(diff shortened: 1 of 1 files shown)


Ganbin:
Nice draft.

My first thinking is about the delay introduced in the fact that when we are the highest bidder, we directly settle and pay the bid before the auction is finished.

This mean that if another higher bid happen, will we be reimbursed directly in the same transaction block or do we have to wait the next transaction block to get our bid back?

What happen if multiple bid arrive in the same block? I suppose the highest win and the others are rejected.

Will this not be too much transactions to be done on L1 in the future when the chain will be busy?

As a transaction block happen every ~52 seconds. Will it not make the process not so reactive as we could have only 1 bid per 52 seconds? Maybe this is enough for an auction. 

Will this CHIP propose a way to delay the auction close every time a new higher bid is placed? This can prevent sniping in the last moment.

joshpainter:
Thank you for your feedback!

> My first thinking is about the delay introduced in the fact that when we are the highest bidder, we directly settle and pay the bid before the auction is finished. 
> 1. This mean that if another higher bid happen, will we be reimbursed directly in the same transaction block or do we have to wait the next transaction block to get our bid back?

In my current thinking, all functionality is executed when a high bidder "spends" the auction singleton. This logic is contained in the PAY_HIGH_BID program. The first high bidder would "spend" the singleton by running PAY_HIGH_BID program, which is going recreate the singleton, now with the high bidder's information embedded. But the singleton will only allow itself to be spent if it sees announcements from all the coins that it knows need to be spent to fulfill the high bid obligations. See CALC_BUYERS_PREMIUM and CALC_COMMISSION. These programs will create the correct coin conditions needed for those features and PAY_HIGH_BID will do the same for the main high bid. In the end, the high bidder will not be able to "steal" (spend) the singleton unless they prove that they have made all the appropriate spends, asserted through announcements.

> 2. What happen if multiple bid arrive in the same block? I suppose the highest win and the others are rejected.

I'm definitely not an expert here, but since high bidders will be competing to spend the same auction singleton, it will be "first come first served." Or more accurately - which spend bundle reached the farmer who eventually farmed that block first? The other one would have been rejected. But with dynamic/auto-extending auctions, this won't be a big deal. More than likely, the DAPP will just try again next block.

> 3. Will this not be too much transactions to be done on L1 in the future when the chain will be busy?
> 4. As a transaction block happen every ~52 seconds. Will it not make the process not so reactive as we could have only 1 bid per 52 seconds? Maybe this is enough for an auction.

The interesting thing about live auctions is it takes months of planning and then they may be over in 60 seconds and 12 bids. This all has to do with the auctioneer of course, but they are trained in the art of high pressure sales and their goal is to get as many bids as fast as possible in order to activate ["auction fever"](https://en.wikipedia.org/wiki/Calor_licitantis) in the bidders. We have no need to do this and much more reason to slow things down and auto-extend times to prevent sniping, better user experience, etc. Modern digital auction users are especially used to these slower-paced auctions.

> 5. Will this CHIP propose a way to delay the auction close every time a new higher bid is placed? This can prevent sniping in the last moment.
Yes! We are still very early. This would be defined in the CALC_HAMMER_TIME program and would probably be the most popular option. Other weird rules could be embedded here during the mint of the auction singleton though! 

The parameters detailed here and even high-level ideas around using singletons are up for debate. I'm certainly not an expert in any of this but not many of us are, so let's start somewhere and figure it out together! :)

joshpainter:
Updated with more community feedback, updates and credit section

iced-caffeine:
Hey guys,

I tried to read through this to figure out where the hangups are.  Were you ever able to figure out if the last bid + blockheight obstacle could be overcome before the high bidder can crack the egg?

Is this actively being worked on at all?

hoffmang9:
Relevant and not available during the initial discussions above, NOT_VALID_AFTER is likely to show up as a CHIP in the next couple of months.

joshpainter:
> I tried to read through this to figure out where the hangups are. Were you ever able to figure out if the last bid + blockheight obstacle could be overcome before the high bidder can crack the egg?

I ended up getting around this by changing the logic, which I think ended up better anyway. From above:

"Actually - are we overthinking this? What is wrong with a ruleset that says "anybody can steal the uncracked egg at any time" plus "the current owner can crack the egg (melt singleton) after the set end date" which will get pushed forward by during the "high bid" spend."
 
This makes it the high-bidder's responsibility to "end" the auction, but is only allowed AFTER a certain blockheight. This fits with the current "Valid after" asserts. This assumes the high-bidder (or their automated dApp) will be incentivized to end the auction ASAP so they can claim their win. But it also gives the high-bidder the weird option of "letting it ride" because of bidder's remorse, and hoping someone else will eventually outbid them, so they get reimbursed for their high bid. Remember that the original seller and even the auction house has already been reimbursed during every bid, so they don't mind if the auction continues indefinitely!

> Is this actively being worked on at all?

Not currently, but it is on my list! And/or others too, hopefully. :)

iced-caffeine:
Thanks for the comments.

In Re: "they don't mind if the auction continues indefinitely!"  This is a spam vector.  Someone can set the block height infinitely high with the first, low bid, thereby effectively burning the asset.  The extend time with a new high bid should be a constant, known param in the program throughout the life of the auction.

iced-caffeine:
Ah, this makes sense.  I misunderstood the initial problem.

Basically after the auction "ends", the high bidder has the right to claim the asset.  This sounds similar to how a solo farmer can claim his 1.75 XCH whenever.  It's there on the chain for them, but they still need to go claim it.

joshpainter:
> Ah, this makes sense. I misunderstood the initial problem.
> 
> Basically after the auction "ends", the high bidder has the right to claim the asset. This sounds similar to how a solo farmer can claim his 1.75 XCH whenever. It's there on the chain for them, but they still need to go claim it.

Exactly!

joshpainter:
> Ah, this makes sense. I misunderstood the initial problem.
> 
> Basically after the auction "ends", the high bidder has the right to claim the asset. This sounds similar to how a solo farmer can claim his 1.75 XCH whenever. It's there on the chain for them, but they still need to go claim it.

Exactly! An interesting side-effect of this is that it inverts auction sniping. Normally, an auction sniper will wait until the last possible moment to make a high-bid. This ends up pushing most of the action to the end of the auction. But in this model, you want to absolutely *make sure* your high bid is in well before the end time, cause you'll want to take into account traffic and confirmation time, etc.

Now, the current high-bidder is the "sniper" in that they are waiting to end that auction *the moment* they can so they can claim their win. The sniper has become the snipee!

PR #25, opened 2022-06-18
Update README.md

CalebCurry:
change H to h


danieljperry:
It was actually supposed to be a capital H because of the acronym.

PR #26, opened 2022-06-25
CHIP-0007: Off-chain metadata format for NFT1

wriches:
A standard for formatting off-chain metadata for NFT1-compliant NFTs on Chia's blockchain


CHIPs/chip-will-riches-metadata.md
```diff
@@ -0,0 +1,101 @@
...
+
+#### Properties
+
+| Property     | Type                    | Required | Description                      |
+|--------------|-------------------------|----------|----------------------------------|
+| `id`         | string                  | **Yes**  | ID of the NFT collection         |
+| `name`       | string                  | **Yes**  | Name of the NFT collection       |
+| `attributes` | [object](#attributes)[] | No       | Attributes of the NFT collection |
+
+#### `attributes`
```
BrandtH22, line 71: See note below, recent notes indicate this should be traits rather than attributes

assets/chip-0007/example.json
```diff
@@ -0,0 +1,48 @@
...
+    "collection": {
+        "name": "Example Pokémon Collection",
+        "id": "e43fcfe6-1d5c-4d6e-82da-5de3aa8b3b57",
+        "attributes": [
+            {
+                "type": "description",
+                "value": "Example Pokémon Collection is the best Pokémon collection. Get yours today!"
+            },
+            {
+                "type": "icon",
```
BrandtH22, line 31: Should we include recommended file type and resolution for the icon?

assets/chip-0007/example.json
```diff
@@ -0,0 +1,48 @@
...
+                "value": "https://examplepokemoncollection.com/image/icon.png"
+            },
+            {
+                "type": "banner",
```
BrandtH22, line 35: Should we include recommended file type and resolution for the banner?

assets/chip-0007/example.json
```diff
@@ -0,0 +1,48 @@
+{
+    "format": "CHIP-0007",
+    "name": "Pikachu",
+    "description": "Electric-type Pokémon with stretchy cheeks",
```
BrandtH22, line 4: I think we need to add a sensitive content object to reduce the amount that marketplaces/explorers need to manually identify:

"sensitive_content": "false"

assets/chip-0007/example.json
```diff
@@ -0,0 +1,48 @@
+{
+    "format": "CHIP-0007",
+    "name": "Pikachu",
+    "description": "Electric-type Pokémon with stretchy cheeks",
+    "minting_tool": "SuperMinter/2.5.2",
+    "attributes": [
```
BrandtH22, line 6: The most recent note I had is that we were changing this from "attributes" to "traits"
This aligns more with the industry standard which uses trait_type rather than just type

assets/chip-0007/example.json
```diff
@@ -0,0 +1,48 @@
...
+            "type": "Friendship",
+            "value": 50,
+            "min_value": 0,
+            "max_value": 255
+        }
+    ],
+    "collection": {
+        "name": "Example Pokémon Collection",
+        "id": "e43fcfe6-1d5c-4d6e-82da-5de3aa8b3b57",
+        "attributes": [
```
BrandtH22, line 25: My notes indicate that we were removing the project information from the collection object and having it listed as indicated below:
```
"collection": {
    "name": "Ninja Turtles",
    "id": "49B83117-3C25-4848-AD9C-A57F320F90C5"
    },
"project": [
     {
       "type": "icon",
       "value": "link to the icon"
     },
     {
       "type": "banner",
       "value": "link to the banner"
     },
     {
       "type": "description",
       "value": "Description text"
     },
     {
       "type": "twitter",
       "value": "twitter handle"
     },
     {
       "type": "website",
       "value": "link to website"
     },
   ],
```

assets/chip-0007/example.json
```diff
@@ -0,0 +1,48 @@
...
+            {
+                "type": "twitter",
+                "value": "ExamplePokemonCollection"
+            },
+            {
+                "type": "website",
+                "value": "https://examplepokemoncollection.com/"
+            }
+        ]
+    }
```
BrandtH22, line 47: Can we add a data object to make this more adaptable for other types of data that projects might include?

"data": {
}

BrandtH22 (commented):
General comments

assets/chip-0007/example.json
wriches: I think this would be best handled outside of the CHIP. There are certainly cases for other types of images ("icon" and "banner" is quite restrictive), so being more prescriptive with just these two types of images has limited value.

I think it makes sense to allow these fields to become standardised more organically, and then a more prescriptive revision of this CHIP could be submitted in the future. This would be submitted under a different CHIP number, so tools would require the logic to recognise that "CHIP-0011 is v2 of CHIP-0007", for example.

assets/chip-0007/example.json
wriches: Same view as for the icon: https://github.com/Chia-Network/chips/pull/26#discussion_r906879164

assets/chip-0007/example.json
wriches: Are there any others than OpenSea that use this? AFAIK they're the only one (albeit they are quite ubiquitous now). Others (e.g. Enjin, Immutable X, Flow) all use a variety of terms.

If there's a desire to align with OpenSea's formatting, then we should change `type` to `trait_type` and leave `attributes` as it is, since this is [consistent with their format](https://docs.opensea.io/docs/metadata-standards#attributes).

assets/chip-0007/example.json
wriches: Could you please elaborate on the reasoning here? I would expect there to always be a description, icon and banner that is specific to the collection, rather than to a project which produced the collection. For things like Twitter and a website link, sometimes these will be specific to the collection, but sometimes they will not.

In the cases where this information is actually the creator's information (e.g. the creator has a website, but the collection itself does not), then it would make more sense for this information to be rooted to the creator's DID rather than included in the NFT's metadata.

assets/chip-0007/example.json
wriches: Seems like a good idea. Could we make it more granular, so that there's the possibility to categorise what aspect of the NFT is sensitive (e.g. language, nudity, horror)?

We could allow an array of objects (without enforcing what these are):

```
"sensitive_content": [
    {
        "language": true,
        "horror": "caution"
    }
]
```

...but also allow a boolean so that it could simply be declared as true or false for now:

```
"sensitive_content": true
```

assets/chip-0007/example.json
wriches: The current dialect allows for this to be included (`"additionalProperties": true`). I can see value in explicitly including a general-purpose data object to provide some guidance, but I wonder if it might be better outside of the collection object (since we probably want to minimise what goes into the collection object for now whilst it's off chain). What do you think?

assets/chip-0007/example.json
paninaro: At a minimum, I agree that supporting `"sensitive_content": true|false` makes sense.

Additionally supporting a list of those categories sounds nice, though I'm not sure what "horror": "caution" means. For simplicity, I would probably just keep it as an array of strings, instead of key/value pairs, e.g. `"sensitive_content": ["drugs", "nudity", "gore"]`, and maybe leave it as a separate exercise to figure out if there needs to be some agreement on what those categories are.

assets/chip-0007/example.json
paninaro: :+1: for `trait_type`

assets/chip-0007/example.json
wriches: > For simplicity, I would probably just keep it as an array of strings

Agreed with this approach. I will update accordingly.

assets/chip-0007/example.json
wriches: Updated from `type` to `trait_type`

assets/chip-0007/example.json
BrandtH22: I can support this also, thank you!

assets/chip-0007/example.json
```diff
@@ -0,0 +1,49 @@
...
+        {
+            "trait_type": "Friendship",
+            "value": 50,
+            "min_value": 0,
+            "max_value": 255
+        }
+    ],
+    "collection": {
+        "name": "Example Pokémon Collection",
+        "id": "e43fcfe6-1d5c-4d6e-82da-5de3aa8b3b57",
```
BrandtH22, line 25: Based on the confusion between series and edition we should add the series information into the collection object here:

"series_number": 1,
"series_total": 10000,

assets/chip-0007/example.json
wriches: The `data` object has been added

assets/chip-0007/example.json
wriches: I've added the `series_number` and `series_total` properties, but I added them at the document root rather than in the `collection` object, since I think the `collection` object should be the same for every NFT within the collection.

DrakoPensulo (commented):
It seems to me that in this code update ("[Add series_number and series_total properties](https://github.com/Chia-Network/chips/pull/26/commits/c9d8a501952ad5cab5857408188bd85bcfc77c46)") the word 'unique' should be replaced with 'distinct'. Justification: There can be a collection which contains 5 distinct NFTs each having some number of editions (identical copies). The NFTs in that collection will be unique only is each distinct NFT has 'total editions = 1'.

wriches:
> It seems to me that in this code update ("[Add series_number and series_total properties](https://github.com/Chia-Network/chips/pull/26/commits/c9d8a501952ad5cab5857408188bd85bcfc77c46)") the word 'unique' should be replaced with 'distinct'.

Agreed, this terminology is clearer in this case. I've updated the wording. Thanks.

assets/chip-0007/example.json
DrakoPensulo: Shouldn't the `series_total` field be moved to the `collection` object? It still guarantees that the `collection` object is the same for every NFT, yet it makes more sense for this field to be there since it is a property of the whole collection.

wriches: `series_total` is not necessarily the same for the whole collection, since you can have multiple series with a single collection. For example, you might have a collection called "Photos of New York", and within that a series of photos of taxis, a series of photos of buildings, etc. Quite often a collection will contain a single series, but not always.

DrakoPensulo: If a collection contains several series/sequences then more than one NFT from that collection may have identical `series_number`. If, additionally `series_total` are equal (for these NFTs) then it is not possible to determine which NFT belongs to which series. I see three possible solutions o distinguish sequences within a collection.
1) Do not allow sequences of the same size within a collection (that is use `series_total`  to distinguish sequences)
2) Introduce a new filed to metadata, `series_id` to distinguish sequences.
3) Recommend creators to use `name` of the NFT in a way which allows series identification. 
It seems to me that solution 2 is the best from the point of view of online services dealing with NFTs.

wriches: Agreed that solution 1 is too restrictive and arbitrary. Solution 2 will not be necessary when collections can be handled in a more verifiable manner, such as through Datalayer. If you introduce a series ID into the off-chain metadata file, there's also an argument to introduce an edition ID. Currently, marketplaces verify the collection by using the ID along with the creator DID, but having to do the same for a series ID (and edition ID) and handle conflicts seems to not be worth the effort. It's worth it for collections, but I expect very few collections will make use of multiple series at this point.

Solution 3 is effectively the status quo (it's common for NFTs to be titled as "Chia Friends #123", for example, both in and outside of the Chia ecosystem) and seems like a sensible approach until something like Datalayer is available for collection and series data.

paninaro:
@rifeljm @wriches what metadata fields should be added to support additional media types? At a minimum I think we should add something like `"preview_uri": "<image thumbnail or video preview uri>` to allow the client to display reasonably-sized (350x350?, 500x500?) artwork associated with the NFT. For audio NFTs this could be the artist's album artwork, or for a video, it could be a preview video. It could also be used for image NFTs to provide a smaller thumbnail.

OR would it be better to have media-specific preview uris? e.g. `preview_image_uri`, `preview_video_uri`. That way, the client wouldn't need to check the content type to determine how to display the preview content.

DrakoPensulo:
I think it will be better to have `"preview_uri"` field since NFT can be a document of any type (image, video, text document, pdf, ...). In order to make content type identification easier to clients,  a field like `"preview_type"` may be added (possibly optional) with values like jpg, png, mov, mp4, pdf, doc, txt, .... etc.

paninaro:
> I think it will be better to have `"preview_uri"` field since NFT can be a document of any type (image, video, text document, pdf, ...). In order to make content type identification easier to clients, a field like `"preview_type"` may be added (possibly optional) with values like jpg, png, mov, mp4, pdf, doc, txt, .... etc.

Yes, if we use a generic `preview_uri` field, we would want to also include a field identifying the type, as you suggested.

I've also been thinking that it would be useful to have a list of preview URIs whose loading priority is expressed in descending order. e.g.:

```
"preview_uris": [
  {
    "uri": <highest priority preview uri>,
    "type": "mp4"
  },
  {
    "uri": <second highest priority preview uri>,
    "type": "png",
  }
]
```

wriches:
> Additionally why the heck is this still an open discussion? This should have been merged and finalezied a month ago. There are over 1k collections and over 130k NFTs out there following this "standard". You missed the chance to adjust stuff now. IMHO if you really want to add/change stuff you should try to define a new CHIP for this.

Because that's within the confines of the CHIP process and would be taken into account for any changes i.e. changes that are breaking for existing NFTs would have a much higher barrier for acceptance.

paninaro:
@speedmann so if I understand correctly, you would suggest that the list of URLs all refer to the same content type? Would the type field then be a sibling of the list?

I'm happy to hear additional thoughts on this as it's in the interest of the community to have this figured out for video/audio/other media types.

paninaro:
> IMHO it should point to the same preview everytime yes. This keeps the behaviour the same as data/metadata/license uris.
> 
> I also would limit the preview to any valid image type as i can really see chances for malicious actors here to annoy the heck out of people (Imagine getting rickrolled by an airdropped NFT)

Agreed that the list should reference the same preview content across multiple mirrors, just like the on-chain URLs.

Re: limiting the preview to images only -- the goal that we're (the Chia frontend team) seeking is to provide artists/creators with a rich representation of their work. While I agree that there will be some actors that seek to abuse the system or annoy others, there are UX solutions to deal with that (don't auto-play videos, hide new NFTs by default, etc.). In fact, my original suggestion of supporting different types of previews was to support apps that may want to disable videos entirely, but still allow image previews.

paninaro:
TL;DR:
Proposing an optional `"content_type": <media type>` field to support displaying NFT content beyond images. Assume image content if the field is not specified.

--

While working on support for displaying additional NFT content types, we've run into a bit of a quandary. Given an arbitrary content URL in the on-chain metadata, how can we determine what the content type is? This impacts how we display that content in the app (i.e. do we render into an `` or `<video>` or `<audio>` tag, etc.)

File extensions are nice when available, but arweave/ipfs links don't provide them (at least on the NFTs that I've been looking at).

Checking the "Content-Type" HTTP header would be great, but again, it's not always available (again, arweave doesn't seem to provide it).

If we naively assume that all NFTs are images, then there's no real issue -- we can simply render into an `` tag and hope for the best.

As we open up to supporting new NFT file types, this has become an issue. And since this CHIP still hasn't been closed, I would like to suggest an optional addition to the off-chain metadata that would support displaying these new NFT file types.

`"content_type": <media type>` where the media type is a standard HTTP media type as described in [RFC 7231](https://www.rfc-editor.org/rfc/rfc7231#section-3.1.1.1)

Examples:
`"content_type": "video/mp4"` --> MPEG-4 video
`"content_type": "audio/ogg"` --> OGG audio track
`"content_type": "application/pdf"` --> PDF document

Using the standard media type format used by the "Content-Type" HTTP header allows implementations to follow existing, well-understood semantics that would normally apply to the HTTP header.

The absence of the field would indicate that the NFT content should be assumed to be an image (as it is today). Since the field would be optional, the display of existing image NFTs minted today would not be impacted.

danieljperry (approved):

PR #30, opened 2022-09-05
CHIP-0008: Splitter Puzzle

greimela:
This proposal aims to define a standard for splitting royalties for NFT1-compliant NFTs, using a standard Chialisp puzzle to prevent marketplace lock-in.

The implementation of the Chialisp is already included, but it has not been audited by any third party yet.

Feedback is very welcome!

### DIFF
--- a/CHIPs/chip-0008.md
+++ b/CHIPs/chip-0008.md
@@ -123,13 +123,47 @@ Using the above scheme as an example, it looks like this:
 ((0x3816a97150946c7745457f45a4fed6c16890637847b2df95f383e439758732d7 80) (0x09731fc56bff8031dc9a16f60dcb6f6462960cd76d4b5233dd5bc21697d7fca7 20)
 ```
 
-The only solution parameter needed on spend is the actual amount of the coin.
+The only solution parameter needed on spend is the actual amount of the coin and the total number of shares. The total number of shares is defined as the sum of all individual shares in the scheme.
 
-On spend, the puzzle steps through the entries of the `PAYOUT_SCHEME` and creates a new coin per entry. The amount of the new coins will be a share of the total amount, determined by the share defined in the `PAYOUT_SCHEME`. The total number of shares is defined as the  sum of all individual shares in the scheme and will be calculated by the puzzle.
+The puzzle is very flexible and doesn't require a specific number of shares.
+This means the following share distributions are equivalent:
 
-This means the share distributions `(1,1)` and `(50,50)` and `(5000, 5000)` are equivalent.
+| Recipient Address                                              | Share      |
+| :------------------------------------------------------------- | :--------- |
+| xch18qt2ju2sj3k8w3290az6flkkc95fqcmcg7edl90ns0jrjav8xttsfmtqfp | 4          |
+| xch1p9e3l3ttl7qrrhy6zmmqmjm0v33fvrxhd494yv7at0ppd97hljns4uw464 | 1          |
+
+is equivalent to
+
+| Recipient Address                                              | Share      |
+| :------------------------------------------------------------- | :--------- |
+| xch18qt2ju2sj3k8w3290az6flkkc95fqcmcg7edl90ns0jrjav8xttsfmtqfp | 80         |
+| xch1p9e3l3ttl7qrrhy6zmmqmjm0v33fvrxhd494yv7at0ppd97hljns4uw464 | 20         |
+
+is equivalent to
+
+| Recipient Address                                              | Share      |
+| :------------------------------------------------------------- | :--------- |
+| xch18qt2ju2sj3k8w3290az6flkkc95fqcmcg7edl90ns0jrjav8xttsfmtqfp | 8000       |
+| xch1p9e3l3ttl7qrrhy6zmmqmjm0v33fvrxhd494yv7at0ppd97hljns4uw464 | 2000       |
+
+This gives the creator of the puzzle a lot of flexibility on how to split the coin.
+
+On spend, the puzzle steps through the entries of the `PAYOUT_SCHEME` and creates a new coin per entry. The amount of the new coins will be a share of the total amount, determined by the share defined in the `PAYOUT_SCHEME` and the total number of shares. 
 
 If the created coins do not use up the full coin amount, the rest is added to the last recipient in the payout scheme. This difference is at most NUMBER_OF_RECIPIENTS mojos, therefore its not worth the effort to construct a more complicated mechanism. It is important that the full coin amount is being reissued, otherwise it breaks CAT coins.
+
+Here are some example coin splits for the above schema
+
+| Coin Amount | Recipient 1 receives | Recipient 2 receives |
+| :---------- | :------------------- | :------------------- |
+| 1 mojos     | 0 mojos              | 1 mojos              |
+| 2 mojos     | 1 mojos              | 1 mojos              |
+| 100 mojos   | 80 mojos             | 20 mojos             |
+| 101 mojos   | 80 mojos             | 21 mojos             |
+| 102 mojos   | 81 mojos             | 21 mojos             |
+| 1000 mojos  | 800 mojos            | 200 mojos            |
+| 1001 mojos  | 800 mojos            | 201 mojos            |
 
 ## Test Cases
 
@@ -150,6 +184,10 @@ Potential risks:
 
 - Royalty recipient can be altered
   - The royalty recipients are curried into the puzzle and the only parameter in the solution is the amount. So the payout scheme can not be altered after it has been created.
+- Puzzle can be bricked by wrong configuration
+  - Due to the dynamic share calculation, any payout scheme with positive integers as shares will work fine.
+  - The nature of Chialisp is that the puzzle will only execute and check the validity of the payout scheme when spending the coin. This is too late to detect any configuration errors, so the tools constructing the puzzle have to make sure that the payout scheme adheres to the defined format.
+  - The puzzle does not support negative shares or fractional shares and will behave in undefined ways if configured that way.
 - Royalty payout can be stalled
   - Since the royalty payment can be triggered by any party on a permissionless blockchain, nobody can stall or prevent this. 
 - Info to construct split puzzle can get lost
--- a/CHIPs/chip-0008.md
+++ b/CHIPs/chip-0008.md
@@ -5,7 +5,7 @@
 | Author       | [Andreas Greimel](https://github.com/greimela) ([@acevail\_](https://twitter.com/acevail_) on Twitter) |
 | Editor       | [Dan Perry](https://github.com/danieljperry)                                                           |
 | Comments-URI | https://github.com/Chia-Network/chips/pull/30                                                          |
-| Status       | Last Call                                                                                                  |
+| Status       | Final                                                                                                  |
 | Category     | Informational                                                                                          |
 | Sub-Category | Informative                                                                                            |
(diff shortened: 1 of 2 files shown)


BrandtH22 (commented):
I might be missing these components but I am not seeing where:
1. Excess mojo are allocated (if the split leaves change what happens to the remainder?)
2. CATs are being handled (likely needs to be done in a driver program and handled per CAT or via a connected API)

Here is another version of this royalty split that takes both of the above into account.
This version is undergoing final testing but is close to production ready:
https://github.com/trgarrett/chialisp/tree/main/royalty_share/clsp/p2_royalty_share_allocated_percentages_feeless

greimela:
Thank you for your review!

Regarding 1: 
It is left to the farmer. The proposal states
> If the created coins do not use up the full coin amount, the rest is left for the farmer of the block. This difference is most likely very small and not worth the effort to construct a more complex puzzle to collect this rest.

Regarding 2:
That is a good point, I will provide a unit test to show how the CAT spend would look like.

Also, thanks for pointing out the other royalty puzzle!
I knew there was one out there, but I couldn't find it again...

@trgarrett, I would love to incorporate your learnings into this proposal!

L4bb3rs:
Been playing with Tgarretts and discussing. Some very good work has been done!

trgarrett:
CATs were the biggest challenge I ran into with my puzzle. Unlike XCH which smoothly flowed to the farmer (other than one potential opportunistic attack due to not using a 52 RESERVE_FEE condition), excess CAT mojos would always break the CAT spend accounting and make the whole coin held by the royalty puzzle unspendable.

Also, because the puzzle address changes with CATs, I had to code in the set of CATs I was willing to support. An API integration or CSV import would work just as well, but would progressively slow the spending down if you had a significant number of them you weren't actually needing to spend. I went this way rather than relying on address hints because, although I was told they were a best practice, they didn't seem to be used consistently in all wallets and reference documentation and I wanted to avoid unspendable (unfindable) coins.

trgarrett:
One other point of feedback that I saw on Twitter, that I had also wondered about was whether this was fully in keeping with the spirit of CHIPs. I do think it would be a great feature to add to one or more Chia clients, but I wouldn't want all the clients/wallets to have an enormous set of features they all had to support in order to be "real" Chia wallets.

From a user interface perspective in the Chia wallet, it would be fairly straightforward to have the mint take parameters like -ra{n}:{address n} -rp{n}basis points and have the royalty puzzle built on the fly. I think the very first feature request you would see if Chia went down that road would be to have a singleton fronting the royalty puzzle to allow the minter to redefine the allocations over time. You would likely end up with all sorts of weird business requirements for that around multisigs, change of ownership, voting on membership allocations, etc., which is where it really starts to sound like something the Chia standards might not want to concern themselves with.

greimela:
> CATs were the biggest challenge I ran into with my puzzle. Unlike XCH which smoothly flowed to the farmer (other than one potential opportunistic attack due to not using a 52 RESERVE_FEE condition), excess CAT mojos would always break the CAT spend accounting and make the whole coin held by the royalty puzzle unspendable.

Good point, I didn't consider that CATs can not be left for the farmer.
So I guess we have to determine the amount left and send it to someone.
I saw your solution with the tip jar, but I think I'd prefer it to go to some address in the list, to simplify things.
The tip jar is a nice idea, but the value will most likely be very small, so I don't think it's work the effort.

> Also, because the puzzle address changes with CATs, I had to code in the set of CATs I was willing to support. An API integration or CSV import would work just as well, but would progressively slow the spending down if you had a significant number of them you weren't actually needing to spend. I went this way rather than relying on address hints because, although I was told they were a best practice, they didn't seem to be used consistently in all wallets and reference documentation and I wanted to avoid unspendable (unfindable) coins.

I only considered the case of NFT royalties, since it seems to be the most popular use case for split puzzles.
NFT generally only pay out royalties when using offers.
In this case, it is actually pretty easy to determine the CAT TAILs.
You just uncurry all the relevant coins of an offer and it will give you all the TAILS involved.

This would not detect the case where somebody wants to send funds to the split puzzle as a transfer, without an offer being involved.
But I would consider that a different use case, which happens to use the same puzzle.

> One other point of feedback that I saw on Twitter, that I had also wondered about was whether this was fully in keeping with the spirit of CHIPs. I do think it would be a great feature to add to one or more Chia clients, but I wouldn't want all the clients/wallets to have an enormous set of features they all had to support in order to be "real" Chia wallets.

I can see that point.
As an alternative to the wallet, the split could also be executed by a minting tool provider or some other player in the ecosystem.
And since the puzzle is open source, it can also be done manually by using an open-source driver like you have provided.

But I don't think the manual split is a reasonable approach for a typical end user.
Some user of the Goby wallet will not know how to set up a Python routine on a server to split her royalties.

trgarrett:
> I saw your solution with the tip jar, but I think I'd prefer it to go to some address in the list, to simplify things.

That’s reasonable. I thought of that. I couldn’t think of a fair way to do that which also minimized computation (fake/derived randomness) so I landed on the tip jar. You could just as well always give it to the highest or lowest earner, I suppose.

CHIPs/chip-XXXX.md
```diff
@@ -0,0 +1,171 @@
...
+    ["xch1p9e3l3ttl7qrrhy6zmmqmjm0v33fvrxhd494yv7at0ppd97hljnscmfrmx", 2000] // 20%
+  ]
+}
+```
+
+The wallet or tool accepting this payout scheme from a user and passing it into the NFT has to ensure the percentages add up to 100%.
+
+## Test Cases
+
+The unit test for the royalty split Chialisp puzzle is located in the `royalty_split` branch of the `greimela/chia-blockchain` GitHub repository: [tests/wallet/nft_wallet/test_nft_puzzles.py](https://github.com/greimela/chia-blockchain/blob/royalty_split/tests/wallet/nft_wallet/test_nft_puzzles.py).
```
trepca, line 141: it'd be great to see some negative tests (empty scheme, wrong `my_mount`, scheme above 100% etc)

CHIPs/chip-XXXX.md
```diff
@@ -0,0 +1,171 @@
...
+
+  ; main
+  (c
+    (list ASSERT_MY_AMOUNT my_amount)
+    (split_amount_and_create_coins PAYOUT_SCHEME my_amount)
+  )
+)
+```
+
+The royalty payout scheme `PAYOUT_SCHEME` is curried into this puzzle, to make it immutable. The only solution parameter needed on spend is the actual amount of the coin.
```
trepca, line 103: it'd be useful to define the structure of `PAYOUT_SCHEME` here

CHIPs/chip-XXXX.md
```diff
@@ -0,0 +1,171 @@
...
+
+On spend, the puzzle steps through the entries of the `PAYOUT_SCHEME` and creates a new coin per entry. The amount of the new coins will be a share of the total amount, determined by the share defined in the `PAYOUT_SCHEME`. The payout share is specified in the same format as NFT royalties, which is `int(percentage * 100)`.
+
+If the created coins do not use up the full coin amount, the rest is left for the farmer of the block. This difference is most likely very small and not worth the effort to construct a more complex puzzle to collect this rest.
```
trepca, line 107: can we define `likely very small` a bit further?

CHIPs/chip-XXXX.md
```diff
@@ -0,0 +1,171 @@
...
+
+The royalty puzzle does not check whether the shares in the payout table add up to 100%, since it would be too late. This verification has to be done by the entity creating the NFT and the payout scheme.
+
+### NFT Metadata
+
+The second part is an addition to the NFT metadata.
+In order to allow any observer to recreate the split royalty puzzle and trigger the split, the observer has to know the payout scheme.
+
+The suggested format is a list of 2-tuples, each containing an address and the percentage of the royalties it should receive, multiplied by 100.
```
CommanderMoto, line 116: This worries me, because it appears to rely on users being able to add up to 100, and no more. A split which added up to >100 seems like it would result in an un-spendable royalty coin?

CHIPs/chip-XXXX.md
```diff
@@ -0,0 +1,171 @@
...
+{
+  //existing metadata
+  "rs": [
+    ["xch18qt2ju2sj3k8w3290az6flkkc95fqcmcg7edl90ns0jrjav8xttsyuvkgj", 8000], // 80%
+    ["xch1p9e3l3ttl7qrrhy6zmmqmjm0v33fvrxhd494yv7at0ppd97hljnscmfrmx", 2000] // 20%
+  ]
+}
+```
+
+The wallet or tool accepting this payout scheme from a user and passing it into the NFT has to ensure the percentages add up to 100%.
```
CommanderMoto, line 137: (Yep, more opportunities for user / software error, this makes me uncomfortable)

CommanderMoto (commented):
Why do all the shares need to add up to 100? Why not just add up the total number of "shares" in the royalty scheme, then apportion payouts based on *that* denominator?

CHIPs/chip-XXXX.md
Rigidity: It would be interesting if it were ratio-based rather than percentages.

For example, to indicate a 50/50 split you could say

A 50
B 50

_Or_

A 1
B 1

This might complicate the puzzle though

CHIPs/chip-XXXX.md
greimela: If I am thinking this through correctly, it should be a maximum of one mojo.
Each `divmod` leaves a value between 0 and 1 mojo behind.
And the sum of all remainders should sum up to 1 mojo.

greimela:
Thanks for the feedback @CommanderMoto @Rigidity!
I have updated the puzzle to derive the number of shares by summing them up.
This should reduce the error potential quite a bit and didn't add a lot of complexity to the puzzle.

CHIPs/chip-0008.md
```diff
@@ -0,0 +1,187 @@
...
+          (get_puzhash (f PAYOUT_SCHEME))
+          (get_amount (f PAYOUT_SCHEME) total_amount total_shares)
+        )
+        (split_amount_and_create_coins (r PAYOUT_SCHEME) total_amount total_shares)
+      )
+      ()
+    )
+  )
+
+  (defun count_shares (PAYOUT_SCHEME)
```
trepca, line 94: can you add a comment why we want to count shares here

trgarrett:
Am I right to think this puzzle doesn't include memos/hints for the wallet? I was experimenting with the puzzle, and I think I got a send to go through. The wallet didn't recognize it in my transaction history though.

trgarrett:
I did some further testing and would recommend including the memo for best results. Shown here: https://github.com/trgarrett/chialisp/pull/3/files/a38dbe55e892203acd8001df974939e74d77f225#diff-6e7343f80d63a39868837db9e25ac65bd69f332b56b24cfef7e18f34c19d9613R39

greimela:
@trgarrett I have update the Chialisp in this proposal to represent our latest changes, including the memo.

hoffmang9 (dismissed):
lgtm

danieljperry:
This CHIP has been reviewed internally at CNI and is now being moved to `Last Call`. This is your last chance to request any changes. If no changes need to be made to the CHIP over the next two weeks, it will be set to `Final`.

danieljperry (approved):
This CHIP is now final. No further changes are allowed, other than errata. Thanks for submitting it, @greimela!

PR #35, opened 2022-10-28
CHIP-0009: Name Service Wallet Address Resolution

danieljperry:


cameroncooper:
Thank you for your time, effort and thoughtfulness in creating this CHIP. In order to maximize efficiency and avoid confusion, I've collated feedback from various interested parties and would like to present it here. Our objectives here are:

1) Ensure that this is something we could implement in Chia’s wallet
2) This CHIP will allow us to support multiple name services in the Chia wallet

Here are a list of questions that came up:

1) There is no specification of encoding or character limitations. Are emoji allowed? Is there any length limit? It would be beneficial to be explicit on what is allowed in a name.

2) The proposal uses the metadata URI list to store the name and expiry. This could lead to invalid metadata URIs if these URIs do not resolve to valid JSON metadata. The CHIP should require the minter to ensure that these additional metadata URIs return valid metadata matching the hash.

3) The NFT1 spec already allows for arbitrary on-chain metadata. This would be the natural place to put new on-chain metadata (see https://github.com/Chia-Network/chia-blockchain/blob/main/chia/rpc/wallet_rpc_api.py#L2115-L2122). Can you move it here instead?

4) This CHIP should provide an API spec for Secondary Resolution Services. Without this many wallets may not be able to support this CHIP, so it would be nice to include a standard here.

5) The CHIP could benefit from some comments or consideration on usability and security issues around typo squatting, name squatting, trademarks, etc. Are there any requirements or guidelines for the minter?

6) If this is to be a generic standard it might be nice to remove the reference to Namesdao here: Thousands of Names have been registered through Namesdao.

7) Is the intention that a wallet will only use a single Name Service DID? Or can multiple Name Service DIDs mint names for a single extension. If a wallet user is expected to only use a single Name Service DID, please provide some consideration for potential pitfalls. For example, if a user wants to send coins to a name, how do they know what Name Service DID to use. What happens if they use the wrong one? Conversely if multiple Name Service DIDs can be used to register under an extension, what is the protocol for determining which NFT should be used for resolution.

One last suggestion is that you might be able to refine the title of the CHIP. This CHIP primarily describes a mechanism for storing and looking up names and addresses in NFTs. It might be nice to include something about the NFT nature the title. Something along the lines of Name Address Resolution Using NFTs. Of course this is your CHIP so I’ll leave the wording up to you if you decide to make a change.

Thanks again for putting this CHIP together. I'm looking forward to seeing this get finalized.

RightSexyOrc:
Thanks for the great feedback. We are going through it in detail, and expect to revise the proposed CHIP with some changes.

RightSexyOrc:
Thanks for these great suggestions and questions. We will answer the first question in some later time. Here are comments regarding the other points.

Q2: The proposal uses the metadata URI list to store the name and expiry. This could lead to invalid metadata URIs if these URIs do not resolve to valid JSON metadata. The CHIP should require the minter to ensure that these additional metadata URIs return valid metadata matching the hash.

A2: Excellent point, agreed and noted

Q3: The NFT1 spec already allows for arbitrary on-chain metadata. This would be the natural place to put new on-chain metadata (see https://github.com/Chia-Network/chia-blockchain/blob/main/chia/rpc/wallet_rpc_api.py#L2115-L2122). Can you move it here instead?

A3: We could see doing this, with a transitionary period for the current metadata uri usage to phase out with the existing expiration of names currently using it over the next year or so, provided that the RPC API would add support for either the 2 new fields (na = name, nx = expiry block?) or else a generalized way to support additional metadata name-key pairs. The RPC API endpoints to update/review would be nft_get_info, nft_mint_nft and nft_mint_bulk.

Q4: This CHIP should provide an API spec for Secondary Resolution Services. Without this many wallets may not be able to support this CHIP, so it would be nice to include a standard here.

A4: Good idea. Rough API spec, will be formalized in updated spec:

Required fields:
name,
address,
nft_coin_id,
updated_block,
expiry_block

A Secondary Resolution Service should publish its API endpoint(s). A service that supports unlimited or very long length Names may need to make a HTTP POST endpoint available for long Names, not only a HTTP GET endpoint. A Secondary Resolution Service should consider potential support for CDN usage and compressed replies, e.g. gzip.

Q5: The CHIP could benefit from some comments or consideration on usability and security issues around typo squatting, name squatting, trademarks, etc. Are there any requirements or guidelines for the minter?

A5: This would generally seem to be out of scope for this CHIP, as it relates to name registration rather than resolution.

Q6: If this is to be a generic standard it might be nice to remove the reference to Namesdao here: Thousands of Names have been registered through Namesdao.

A6: ok

Q7: Is the intention that a wallet will only use a single Name Service DID? Or can multiple Name Service DIDs mint names for a single extension. If a wallet user is expected to only use a single Name Service DID, please provide some consideration for potential pitfalls. For example, if a user wants to send coins to a name, how do they know what Name Service DID to use. What happens if they use the wrong one? Conversely if multiple Name Service DIDs can be used to register under an extension, what is the protocol for determining which NFT should be used for resolution.

A7: It is the expectation that a wallet will use no more than one single Name Service per extension. By extension, we refer to a potential name ending after a period, e.g. .eth or .xch or .btc. A wallet may also have a default Name Service. Use of more than one registry per extension risks different registrars issuing registration of the same name to different addresses, significant user confusion and related pitfalls. Looking at the history of different name services, DNS, ICANN, etc., this approach of one service per extension appears to be reasonable. The potential risk of one Name Service per extension is the dependency of users upon one registrar for the names. In the case of centralized registration issuance, there's a risk that the registry may have issues including even a compromise of its registration key, or government seizure and compromise of the service in the case of a centralized service from a doxxed team. In the case of a Name Service that is decentralized based on a SmartCoin system, there are security risks that the code may have bugs. At the same time, in the case of such compromises the community could potentially choose to migrate to a new registrar for the extension, using a snapshot of the namespace prior to the compromise to re-issue names under a new DID.


Q8: re include NFT mention in title of CHIP

A8: sure

RightSexyOrc:
Regarding the third question, we will proceed with the current design rather than changing the in-NFT metadata storage at this time. At some point in the future, when RPC API endpoints get updated to easily support additional in-NFT metadata, we can look at updating with a new spec and a one year plus transition period, which would be required anyway. 

I have updated my local copy with changes to the CHIP and will be reviewing in the next few days, and then updating to this repository: 

https://github.com/theNamesdao/chips/commit/58f55e0d168236a7bbe4446e59b10b3a0a99ff69

danieljperry:
This CHIP is now in `Last Call`. If no changes are required in the next two weeks, it will be finalized.

danieljperry:
This Chip is now `Final`. It may no longer be modified, other than by adding errata. Thanks @RightSexyOrc for creating it!

hoffmang9 (approved):
lgtm

PR #37, opened 2022-11-03
added link to CHIP-0002 Q&A video

roybotbot:


danieljperry (changes requested):
This CHIP has been finalized, so only errata can be added. It's fine to add this link, but instead of its current location, can you create a new section at the end titled `Errata` and add it there?

danieljperry (approved):
lgtm

PR #46, opened 2022-12-20
CHIP-0011: BLS/SECP CLVM Operators and SOFTFORK Condition

danieljperry:
Opening PR for consideration of the BLS Additions CHIP.

### DIFF
--- a/CHIPs/chip-0011.md
+++ b/CHIPs/chip-0011.md
@@ -390,9 +390,9 @@ An example of using the `softfork` operator to call the new `coinid` operator is
   (q . ()))   ; environment to softfork
 ```
 
-### Hard Fork Addition
+### Hard Fork Additions
 
-The following condition will become available after the hard fork from [CHIP-12](https://github.com/Chia-Network/chips/pull/53) has activated.
+The following condition and optimizations will become available after the hard fork from [CHIP-12](https://github.com/Chia-Network/chips/pull/53) has activated.
 
 #### SOFTFORK
 
@@ -405,6 +405,19 @@ Arguments: The cost of the condition (specified in ten-thousands) is required as
 Usage: `(90 cost)`
 
 CLVM Cost: Varies, specified during the call, can be zero
+
+#### Block Generator Optimizations
+
+As part of the block creation process, a CLVM program called the _ROM generator_ is responsible for the following:
+* Calling the block generator, which outputs all spends, puzzles and solutions for that block
+* Running the puzzles with each corresponding solution
+* Computing puzzle hashes, which are later used to find the coins to spend from the coin set
+
+Prior to the hard fork from CHIP-12, the cost of running the ROM generator counts against the per-block cost limit.
+
+After the hard fork has activated, the ROM generator will no longer be charged a cost. Note that the block generator, puzzles, and solutions will continue to incur the same costs as before.
+
+The hard fork will also enable an updated serialization format for CLVM, which will be used for block generators. The generators will memoize the tree hashes of sub-trees, an optimization that allows a more compact representation of repeated sub-trees. Since identical trees have already been deduplicated by the farmer, it will become cheaper to compute the puzzle hashes in the common case where there are repeated puzzles in a block.
 
 ## Test Cases
 
--- a/CHIPs/chip-0011.md
+++ b/CHIPs/chip-0011.md
@@ -66,8 +66,6 @@ To support the complete set of available BLS operations within the CLVM, math pr
 
 ### bls_g1_add
 
-Opcode: [todo]
-
 Functionality: Add two or more G1 points
 
 Note: This operator is already implemented in CLVM as [`point_add`](https://chialisp.com/operators#bls12-381). For consistency and ease of use, a new compiler macro will be created to map `point_add` to `bls_g1_add` so either name can be used with full backward compatibility.
@@ -80,7 +78,7 @@ CLVM Cost: `101 094` base, `1 343 980` per argument
 
 ### bls_g1_subtract
 
-Opcode: [todo]
+Opcode: 49
 
 Functionality: Subtract one or more G1 points from a base G1 point
 
@@ -92,7 +90,7 @@ CLVM Cost: `132 332` base, `1 362 553` per argument
 
 ### bls_g1_multiply
 
-Opcode: [todo]
+Opcode: 50
 
 Functionality: Multiply a G1 point by a scalar value
 
@@ -104,7 +102,7 @@ CLVM Cost: `2 154 347` base, `12` per byte in the scalar
 
 ### bls_g1_negate
 
-Opcode: [todo]
+Opcode: 51
 
 Functionality: Negate a G1 point
 
@@ -114,9 +112,9 @@ Usage: `(bls_g1_negate point1)`
 
 CLVM Cost: `470 779`
 
-Opcode: [todo]
-
 ### bls_g2_add
+
+Opcode: 52
 
 Functionality: Add two or more G2 points
 
@@ -128,7 +126,7 @@ CLVM Cost: `45 440` base, `5 544 581` per argument
 
 ### bls_g2_subtract
 
-Opcode: [todo]
+Opcode: 53
 
 Functionality: Subtract one or more G2 points from a base G2 point
 
@@ -140,7 +138,7 @@ CLVM Cost: `146 290` base, `5 495 272` per argument
 
 ### bls_g2_multiply
 
-Opcode: [todo]
+Opcode: 54
 
 Functionality: Multiply a G2 point by a scalar value
 
@@ -152,7 +150,7 @@ CLVM Cost: `10 078 145` base, `12` per byte in the scalar
 
 ### bls_g2_negate
 
-Opcode: [todo]
+Opcode: 55
 
 Functionality: Negate a G2 point
 
@@ -164,7 +162,7 @@ CLVM Cost: `1 881 699`
 
 ### bls_gt_add
 
-Opcode: [todo]
+Opcode: 56
 
 Functionality: Add two or more Gt points
 
@@ -176,7 +174,7 @@ CLVM Cost: `60 118` base, `62 655 353` per argument
 
 ### bls_gt_subtract
 
-Opcode: [todo]
+Opcode: 57
 
 Functionality: Subtract one or more Gt points from a base Gt point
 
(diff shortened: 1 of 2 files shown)


hoffmang9 (dismissed):
lgtm

freddiecoleman (approved):
lgtm

hoffmang9 (dismissed):
lgtm. Shall I merge?

danieljperry:
No, let's keep it open while the CHIP is active. People can post their reviews as comments here.

danieljperry:
This video explains the details of this CHIP, including why we would like to add these operators and the technical details of how it would work:
https://www.youtube.com/watch?v=c4tKmjJEJCI
Reviews welcome!

danieljperry:
This CHIP has been updated to include a tentative soft fork block height. It will be moved to `Review` after the operators are confirmed included.

danieljperry:
This CHIP is now in `Review`. Note that a few of the operators have been added, removed, and modified since the CHIP was created. Significant changes are unlikely going forward. Please take another look at the CHIP's contents and leave your reviews here.

danieljperry:
The ability to verify secp256k1 and secp256r1 are now included with this CHIP. Please take a moment to review these additions and leave your thoughts here.

CHIPs/chip-0011.md
```diff
@@ -0,0 +1,411 @@
...
+
+#### bls_verify
+
+Opcode: 59
+
+Functionality: Validates the messages (`msg`) given their public key (`G1`) against the signature (`G2`). Returns `nil` if the signature is valid, otherwise raise an exception. The validation uses the Augmented scheme, which means the G1 points are prepended to the messages before being hashed to G2 points. The DST is `BLS_SIG_BLS12381G2_XMD:SHA-256_SSWU_RO_AUG_`. The validation includes the pair of the negated G1 generator and the signature in the underlying pairing operation
+
+Arguments: A G2 point followed by a list of G1/msg pairs
+
+Usage: `(bls_verify g1point1 msg1 g1point2 msg2 … g1pointn msgn)`
```
SlowestTimelord, line 228: Is this usage example missing a G2 point as the first argument?

CHIPs/chip-0011.md
```diff
@@ -0,0 +1,411 @@
...
+
+- `pubkey` -- the secp256k1 public key used for signing
+- `msg_digest` -- the sha256 hash of the message that was signed
+- `signature` -- the digital signature to verify
+
+Note about `msg_digest` (the sha256 hash of the message to be signed): this parameter is different from `bls_verify`, which has `hash-to-curve` (or `g2_map`) built-in. It simplifies the cost model, since all parameters are nearly fixed in size.
+
+Usage: `(secp256k1_verify pubkey msg_digest signature)`
+
+CLVM Cost: `850 000`
```
SlowestTimelord, line 296: Why is k1 cost so much cheaper than r1? Are these costs benchmarked or from a cost breakdown like other opcodes above?

CHIPs/chip-0011.md
```diff
@@ -0,0 +1,411 @@
...
+
+The following condition will become available after the hard fork from [CHIP-12](https://github.com/Chia-Network/chips/pull/53) has activated.
+
+#### SOFTFORK
+
+Opcode: 90
+
+Functionality: Allows future conditions with non-zero CLVM costs to be added as soft forks. (This functionality was previously only possible as a hard fork)
+
+Arguments: The cost of the condition (specified in ten-thousands) is required as the first argument, further arguments are unspecified. The reason to scale the cost by `10 000` is to make the argument smaller. For example, a `cost` of `100` in this condition would equate to an actual cost of 1 million (`1 000 000`)
```
SlowestTimelord, line 369: How does a low cost (e.g. `coinid`) get specified?

CHIPs/chip-0011.md
```diff
@@ -0,0 +1,411 @@
...
+## Test Cases
+
+The test cases for this CHIP are located in the [op-tests](https://github.com/Chia-Network/clvm_rs/tree/main/op-tests) folder of the clvm_rs GitHub repository. These tests include:
+* [test-bls-ops.txt](https://github.com/Chia-Network/clvm_rs/tree/main/op-tests/test-bls-ops.txt)
+* [test-blspy-g1.txt](https://github.com/Chia-Network/clvm_rs/tree/main/op-tests/test-blspy-g1.txt)
+* [test-blspy-g2.txt](https://github.com/Chia-Network/clvm_rs/tree/main/op-tests/test-blspy-g2.txt)
+* [test-blspy-hash.txt](https://github.com/Chia-Network/clvm_rs/tree/main/op-tests/test-blspy-hash.txt)
+* [test-blspy-pairing.txt](https://github.com/Chia-Network/clvm_rs/tree/main/op-tests/test-blspy-pairing.txt)
+* [test-blspy-verify.txt](https://github.com/Chia-Network/clvm_rs/tree/main/op-tests/test-blspy-verify.txt)
+* [test-more-ops.txt](https://github.com/Chia-Network/clvm_rs/blob/main/op-tests/test-more-ops.txt)
```
SlowestTimelord, line 384: Missing
https://github.com/Chia-Network/clvm_rs/blob/main/op-tests/test-secp256k1.txt
https://github.com/Chia-Network/clvm_rs/blob/main/op-tests/test-secp256r1.txt

SlowestTimelord:
Reading about potential security concerns with how the secp256r1 constants were chosen, is there a particular reason to include it at all in addition to secp256k1?

CHIPs/chip-0011.md
hoffmang9: Out of Bitcoin Core work, there have been very significant performance increases developed for k1 that have not occurred in r1. Hence this is a pretty empirical comparison of the two operators using the rust implementations of both.

hoffmang9:
> Reading about potential security concerns with how the secp256r1 constants were chosen, is there a particular reason to include it at all in addition to secp256k1?

So there is certainly concern among some that the NSA may have poisoned r1's constants. I personally am dubious because it's the primary go to for Apple and Samsung and others in their Secure Elements. The "pro" here though is that r1 is NIST certified while k1 is not.

Choice of curves will remain fully optional for developers and end users also.

SlowestTimelord:
> So there is certainly concern among some that the NSA may have poisoned r1's constants. I personally am dubious because it's the primary go to for Apple and Samsung and others in their Secure Elements. The "pro" here though is that r1 is NIST certified while k1 is not.
> 
> Choice of curves will remain fully optional for developers and end users also.

Ah the industry use and NIST certification makes sense especially in the context of the May 2023 AMA. Thanks Gene.

CHIPs/chip-0011.md
danieljperry: added, thanks

CHIPs/chip-0011.md
danieljperry: Ten thousand is the minimum cost.

SlowestTimelord: I'd like to understand the rationale for 10,000 specifically as the scaling factor. Why not 1,000 or 100,000? Are most op codes expected to be on the order of 10,000 cost? What is the impact of not being specific enough?

SlowestTimelord:
> %: aka modulo, (Opcode: 60)
> modpow: computes (base ^ exponent) % modulus, (Opcode: 61)

The CHIP seems to indicates % as 61 and modpow as 60

danieljperry:
Sorry about that. I transposed them in the comment, which I have now fixed. The CHIP itself is correct.

CHIPs/chip-0011.md
```diff
@@ -0,0 +1,453 @@
...
+
+Opcode: 61
+
+Functionality: Computes the remainder from `numerator` divided by `denominator`
+
+Arguments: A `numerator` and a `denominator`
+
+Usage: `(% numerator denominator)`
+
+CLVM Cost: `998` base, `4` per byte (sum of both operands)
```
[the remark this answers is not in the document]
danieljperry: You are correct, good catch.

CHIPs/chip-0011.md
danieljperry: To be clear, these are conditions. (opcodes mostly refer to operators in CLVM). The rationale to have a scaling factor at all is to save space (and cost for bytes) by not having to specify the trailing zeros in the full cost. The scaling factor could probably have been 100'000 or 200'000 as well. But conditions with costs less than 10'000 are not likely.

For comparison, the existing conditions with cost are `CREATE_COIN` at 1'800'000 and `AGG_SIG_*` at 1'200'000

CHIPs/chip-0011.md
```diff
@@ -0,0 +1,453 @@
...
+
+CLVM Cost:
+
+- `base` -- `17 000` base, `38` per byte
+- `exponent` -- `3` per square byte
+- `modulus` -- `21` per square byte
+
+#### %
+
+Opcode: 61
```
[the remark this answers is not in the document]
danieljperry: > Is this just meant to be a simpler (and cheaper) version of what `divmod` already handles?

Correct, this operator only gives the remainder, and the cost is slightly less than that of `divmod`.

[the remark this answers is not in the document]
danieljperry: > Any restrictions on negative numbers with `%`?

Negative numbers are allowed. I have added a clarifying note to the CHIP.

danieljperry:
This CHIP now includes the ability to add new CLVM conditions with cost. This functionality will become available after next year's hard fork. Currently, new conditions must have zero cost.
This CHIP also includes new AGG_SIG conditions, which only contain one or two of `parent coin`, `puzzle hash`, and `amount` in their signed messages (the current signature conditions contain all three). They will also include unique domain strings to prevent replay attacks.

Please review at your convenience.

danieljperry:

danieljperry:
The activation height for the fork described in this CHIP is now set to 4,474,000. Nothing else has been changed. This CHIP will remain in `Last Call` and the roughly two week window will be reset.

danieljperry:
We have to adjust the activation height for this CHIP once again, so I'm pulling it out of Last Call until we can determine a new height. Nothing else has changed with this CHIP.

danieljperry:
The soft fork activation height for this CHIP has been updated to 4,510,000 and the CHIP has been moved back into `Last Call`. If it does not need to be further modified for the next two weeks (roughly), then it will be moved to `Final`.

danieljperry:

Rigidity (approved):

PR #47, opened 2022-12-24
Capitalize CHIP-0001 reference in CHIP-0001

altendky:
https://github.com/Chia-Network/chips/blob/80e4611fe52b174bf1a0382b9dff73805b18b8c6/CHIPs/chip-0001.md#chip-numbers
> When referring to a CHIP by number, it should be written in the hyphenated form CHIP-WXYZ where WXYZ is the CHIP's assigned number.


danieljperry (approved):
lgtm

PR #53, opened 2023-01-20
CHIP-0012: Reduce Plot Filter

jmhands:
plot filter chip, open up at 8am 1/20

### DIFF
--- a/CHIPs/chip-0012.md
+++ b/CHIPs/chip-0012.md
@@ -68,7 +68,6 @@ The real leverage comes in when a plot is created in under 18.75 seconds (realis
 
 | phase 1 time | plot filter | leverage factor (plots) | space spoofed (TiB)      |
 | ------------ | ----------- | ----------------------- | ------------------------ |
-|              |             |                         |                          |
 | 28.125       | 512         | 171                     | 16.9                     |
 | 18.75        | 512         | 512                     | 50.7                     |
 | 9.375        | 512         | 1536                    | 152.0                    |
@@ -192,15 +191,100 @@ We decided to propose a reduction of the plot filter every three years until the
 
 ## Specification
 
-The final specification will be listed here.
+The [PR](https://github.com/Chia-Network/chia-blockchain/pull/15336) that implements this CHIP has the specification details.
+
+Here is a summary from that PR:
+
+The Proof of Space code was originally written with the assumption that the filter size would remain constant. The method for determining whether a plot passes the filter (from [proof_of_space.py](https://github.com/Chia-Network/chia-blockchain/blob/main/chia/types/blockchain_format/proof_of_space.py)) originally contained `ConsensusConstants` in its signature:
+
+```
+def passes_plot_filter(
+    constants: ConsensusConstants,
+    plot_id: bytes32,
+    challenge_hash: bytes32,
+    signage_point: bytes32,
+) -> bool:
+```
+
+This signature was updated to take in a new variable called `prefix_bits`:
+
+```
+def passes_plot_filter(
+    prefix_bits: int,
+    plot_id: bytes32,
+    challenge_hash: bytes32,
+    signage_point: bytes32,
+) -> bool:
+```
+
+`prefix_bits` is calculated from a new method called `calculate_prefix_bits`, which starts with the hard-coded number of bits (9) and subtracts the appropriate number based on the current block height (prior to the first fork point, no bits are subtracted). Each bit that is subtracted will cut the filter in half:
+
+```
+def calculate_prefix_bits(constants: ConsensusConstants, height: uint32) -> int:
+    prefix_bits = constants.NUMBER_ZERO_BITS_PLOT_FILTER
+    if height >= constants.PLOT_FILTER_32_HEIGHT:
+        prefix_bits -= 4
+    elif height >= constants.PLOT_FILTER_64_HEIGHT:
+        prefix_bits -= 3
+    elif height >= constants.PLOT_FILTER_128_HEIGHT:
+        prefix_bits -= 2
+    elif height >= constants.HARD_FORK_HEIGHT:
+        prefix_bits -= 1
+
+    return max(0, prefix_bits)
+```
+
+The signature of `verify_and_get_quality_string` also had to be modified to include `filter_prefix_bits`:
+
+```
+def verify_and_get_quality_string(
+    pos: ProofOfSpace,
+    constants: ConsensusConstants,
+    original_challenge_hash: bytes32,
+    signage_point: bytes32,
+    *,
+    height: Optional[uint32],
+    filter_prefix_bits: Optional[uint8] = None,
+) -> Optional[bytes32]:
+```
+
+In addition, the harvester and farmer protocols had to be made aware of `filter_prefix_bits`.
+
+Harvester:
+
+```
+class NewSignagePointHarvester(Streamable):
+    challenge_hash: bytes32
+    difficulty: uint64
+    sub_slot_iters: uint64
+    signage_point_index: uint8
+    sp_hash: bytes32
+    pool_difficulties: List[PoolDifficulty]
+    filter_prefix_bits: uint8
+```
+
+Farmer:
+
+```
+class NewSignagePoint(Streamable):
+    challenge_hash: bytes32
+    challenge_chain_sp: bytes32
+    reward_chain_sp: bytes32
+    difficulty: uint64
+    sub_slot_iters: uint64
+    signage_point_index: uint8
+    filter_prefix_bits: uint8
+```
 
 ## Test Cases
 
-Tests are currently being written and will be added here for final review.
+The test cases for this CHIP are located in [test_proof_of_space.py](https://github.com/Chia-Network/chia-blockchain/blob/main/tests/core/custom_types/test_proof_of_space.py), as well as in the class `Mode(Enum):` of [conftest.py](https://github.com/Chia-Network/chia-blockchain/blob/main/tests/conftest.py).
 
 ## Reference Implementation
 
-[PR #15299](https://github.com/Chia-Network/chia-blockchain/pull/15299) of the `chia-blockchain` repository implements the infrastructure for this CHIP.
+This CHIP is being implemented in the following PRs from the `chia-blockchain` repository:
+* [15299](https://github.com/Chia-Network/chia-blockchain/pull/15299) -- infrastructure
+* [15336](https://github.com/Chia-Network/chia-blockchain/pull/15336) -- primary implementation
 
 ## Security
 
--- a/CHIPs/chip-0012.md
+++ b/CHIPs/chip-0012.md
@@ -5,7 +5,7 @@ Description   | Every three years, cut the plot filter in half until it reaches 
 Author        | [JM Hands](https://github.com/jmhands)
 Editor        | [Dan Perry](https://github.com/danieljperry)
 Comments-URI  | [CHIPs repo, PR #53](https://github.com/Chia-Network/chips/pull/53)
-Status        | Last Call
+Status        | Final
 Category      | Standards Track
 Sub-Category  | Core
 Created       | 2023-01-18
--- a/CHIPs/chip-0012.md
+++ b/CHIPs/chip-0012.md
@@ -1,11 +1,11 @@
-CHIP Number   | < Creator must leave this blank. Editor will assign a number.>
+CHIP Number   | 0012
 :-------------|:----
 Title         | Decrease Plot Filter
 Description   | Every two years, cut the plot filter in half until it reaches size 32.
 Author        | [JM Hands](https://github.com/jmhands)
 Editor        | [Dan Perry](https://github.com/danieljperry)
-Comments-URI  | < Creator must leave this blank. Editor will assign a URI.>
-Status        | < Creator must leave this blank. Editor will assign a status.>
+Comments-URI  | [CHIPs repo, PR #53](https://github.com/Chia-Network/chips/pull/53)
+Status        | Draft
 Category      | Standards Track
 Sub-Category  | Core
 Created       | 2023-01-18
--- a/CHIPs/chip-0012.md
+++ b/CHIPs/chip-0012.md
@@ -286,6 +286,10 @@ This CHIP is being implemented in the following PRs from the `chia-blockchain` r
 * [15299](https://github.com/Chia-Network/chia-blockchain/pull/15299) -- infrastructure
 * [15336](https://github.com/Chia-Network/chia-blockchain/pull/15336) -- primary implementation
 
+The following `clvm_rs` PRs are also part of the hard fork from this CHIP:
+* [306](https://github.com/Chia-Network/clvm_rs/pull/306) -- add flag `ENABLE_FIXED_DIV`
+* [173](https://github.com/Chia-Network/chia_rs/pull/173) -- all new arguments in `AGG_SIG_*` conditions
+
 ## Security
 
 At each fork point block height specified in this CHIP, the filter size will be cut in half. Because of this, twice as many plots will pass the filter, and twice as many proofs of space will be found. The effect will be a temporary doubling in the speed of the network, which could cause a temporary outage of some nodes.
--- a/CHIPs/chip-0012.md
+++ b/CHIPs/chip-0012.md
@@ -5,7 +5,7 @@ Description   | Every three years, cut the plot filter in half until it reaches 
 Author        | [JM Hands](https://github.com/jmhands)
 Editor        | [Dan Perry](https://github.com/danieljperry)
 Comments-URI  | [CHIPs repo, PR #53](https://github.com/Chia-Network/chips/pull/53)
-Status        | Draft
+Status        | Review
 Category      | Standards Track
 Sub-Category  | Core
 Created       | 2023-01-18
@@ -192,11 +192,11 @@ We decided to propose a reduction of the plot filter every three years until the
 
 ## Specification
 
-[todo]
+The final specification is being written and will be detailed here.
 
 ## Test Cases
 
-[todo]
+Tests are currently being written and will be added here for final review.
 
 ## Reference Implementation
 
--- a/CHIPs/chip-0012.md
+++ b/CHIPs/chip-0012.md
@@ -5,7 +5,7 @@ Description   | Every three years, cut the plot filter in half until it reaches 
 Author        | [JM Hands](https://github.com/jmhands)
 Editor        | [Dan Perry](https://github.com/danieljperry)
 Comments-URI  | [CHIPs repo, PR #53](https://github.com/Chia-Network/chips/pull/53)
-Status        | Review
+Status        | Last Call
 Category      | Standards Track
 Sub-Category  | Core
 Created       | 2023-01-18
--- a/CHIPs/chip-0012.md
+++ b/CHIPs/chip-0012.md
@@ -192,7 +192,7 @@ We decided to propose a reduction of the plot filter every three years until the
 
 ## Specification
 
-The final specification is being written and will be detailed here.
+The final specification will be listed here.
 
 ## Test Cases
 
--- a/CHIPs/chip-0012.md
+++ b/CHIPs/chip-0012.md
@@ -1,7 +1,7 @@
 CHIP Number   | 0012
 :-------------|:----
 Title         | Decrease Plot Filter
-Description   | Every two years, cut the plot filter in half until it reaches size 32.
+Description   | Every three years, cut the plot filter in half until it reaches size 32.
 Author        | [JM Hands](https://github.com/jmhands)
 Editor        | [Dan Perry](https://github.com/danieljperry)
 Comments-URI  | [CHIPs repo, PR #53](https://github.com/Chia-Network/chips/pull/53)
@@ -16,7 +16,7 @@ Superseded-By | None
 
 ## Abstract
 
-In [Chia's consensus](https://docs.chia.net/consensus-intro), the plot filter is a constant, which each plot has a 1/plot filter probability of being eligible for a challenge for a given signage point. The plot filter's primary benefit is a reduction in disk I/O. Its primary tradeoff is an amplification of the effect of plot grinding. Plot grinding will be feasible with multiple GPUs, and if nothing is changed with Chia's consensus or plot structure, it may become economical with future changes in xch price and advances in computing technology. This proposal will mitigate the cost-effectiveness of plot grinding by reducing the size of the plot filter every two years, to `256`, `128`, `64`, and finally `32`.
+In [Chia's consensus](https://docs.chia.net/consensus-intro), the plot filter is a constant, which each plot has a 1/plot filter probability of being eligible for a challenge for a given signage point. The plot filter's primary benefit is a reduction in disk I/O. Its primary tradeoff is an amplification of the effect of plot grinding. Plot grinding will be feasible with multiple GPUs, and if nothing is changed with Chia's consensus or plot structure, it may become economical with future changes in xch price and advances in computing technology. This proposal will mitigate the cost-effectiveness of plot grinding by reducing the size of the plot filter every three years, to `256`, `128`, `64`, and finally `32`.
 
 ## Definitions
 
@@ -111,18 +111,18 @@ Reducing the filter provides a proportional reduction in spoofed space for plot 
 
 ### Proposal
 
-This CHIP is a proposal to decrease the economic benefit of plot grinding. One year after a pre-determined hard fork activation, the size of the plot filter will be reduced to `256`, thereby cutting in half the number of plots a plot grinding computer can mimic. This proposal will continue to cut the filter in half every two years, until the filter size is `32`.
+This CHIP is a proposal to decrease the economic benefit of plot grinding. One year after a pre-determined hard fork activation, the size of the plot filter will be reduced to `256`, thereby cutting in half the number of plots a plot grinding computer can mimic. This proposal will continue to cut the filter in half every three years, until the filter size is `32`.
 
 If this proposal is accepted and is successful, plot grinding will not be economical for the foreseeable future of computing devices. By using this economic incentive to encourage farmers to store their plots, Chia's PoST consensus will continue to be significantly more energy-efficient than blockchains that use PoW.
 
 The specific block heights at which the plot filter will be reduced are as follows:
 
-| Block height | Year (approx) | Filter size |
-|:-----------: | :-----------: | :---------: |
-| [todo]       | 2024          | 256         |
-| [todo]       | 2026          | 128         |
-| [todo]       | 2028          |  64         |
-| [todo]       | 2030          |  32         |
+| Block height | Month/Year (approx) | Filter size |
+|------------: | ------------------: | ----------: |
+|  `5 496 000` | June 2024           | 256         |
+| `10 542 000` | June 2027           | 128         |
+| `15 592 000` | June 2030           |  64         |
+| `20 643 000` | June 2033           |  32         |
 
 ## Backwards Compatibility
 
@@ -188,7 +188,7 @@ Con:
 
 ### Decision: filter reduction
 
-We decided to propose a reduction of the plot filter every two years until the size reaches `32`. An important consideration was whether harvesters could keep up with the network. As demonstrated previously, if the size of the plot filter were set to `1`, spinning disks would not be viable. However, when the plot filter reaches its final level proposed in this CHIP of `32` in 2030, farmers will be able to keep up with the network using minimum spec hardware.
+We decided to propose a reduction of the plot filter every three years until the size reaches `32`. An important consideration was whether harvesters could keep up with the network. As demonstrated previously, if the size of the plot filter were set to `1`, spinning disks would not be viable. However, when the plot filter reaches its final level proposed in this CHIP of `32` in 2030, farmers will be able to keep up with the network using minimum spec hardware.
 
 ## Specification
 
--- a/CHIPs/chip-0012.md
+++ b/CHIPs/chip-0012.md
@@ -200,7 +200,7 @@ Tests are currently being written and will be added here for final review.
 
 ## Reference Implementation
 
-[todo]
+[PR #15299](https://github.com/Chia-Network/chia-blockchain/pull/15299) of the `chia-blockchain` repository implements the infrastructure for this CHIP.
 
 ## Security
 
--- a/CHIPs/chip-plot_filter.md
+++ b/CHIPs/chip-plot_filter.md
@@ -1,0 +1,222 @@
+CHIP Number   | < Creator must leave this blank. Editor will assign a number.>
+:-------------|:----
+Title         | Decrease Plot Filter
+Description   | Every two years, cut the plot filter in half until it reaches size 32.
+Author        | [JM Hands](https://github.com/jmhands)
+Editor        | [Dan Perry](https://github.com/danieljperry)
+Comments-URI  | < Creator must leave this blank. Editor will assign a URI.>
+Status        | < Creator must leave this blank. Editor will assign a status.>
+Category      | Standards Track
+Sub-Category  | Core
+Created       | 2023-01-18
+Requires      | None
+Replaces      | None
+Superseded-By | None
+
+
+## Abstract
+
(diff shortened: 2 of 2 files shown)


SlowestTimelord:
Thanks for writing such a thorough explanation of the problem and proposed solution @jmhands! A few clarifying questions that I have already asked you in other channels, but adding here for completeness:

1. Does reducing the plot filter by half also double plot "decompression" farming workload? Or is it not exactly proportional?
2. Upon reaching size 32 plot filter, does that imply increasing minimum k-size to k33 is the logical next proactive step? At that time would plot filter reset to 512 or stay at 32?
3. Is increasing the minimum k-size a soft fork?
4. In the event of a hard fork, would that be an opportunity to introduce other pending changes that require a hard fork? What such changes might those be?
5. Would increasing the number of signage points per subslot (to decrease the time between infusion points) be a viable option as well or is that too drastic a change to the consensus?

SlowestTimelord:
A few more thoughts to stir the pot in favor of Option 2 (k-size change):

- If there were ever a time to declare timelines for a minimum k-size, it's probably now while folks are considering replotting anyway.
- I don't think a proposal on plot filter decrease is complete without considering when/how minimum k-size requirement might fit into the ~10 year timeline.
- Plot filter changes + decompression considerations adds a dimension that many farmers may have trouble grasping or considering whereas a minimum k-size timeline is easier to grasp (say, k33 phases in 2025, k34 phases in 2028 etc.)
- Given the unpredictability of future hardware and plotting/farming advances, there's a chance another hard fork to change the timeline of plot filter changes may be necessary whereas as Bram has mentioned in the past a minimum a k-size change would be a soft fork (I'd like to confirm whether or not this is the case).

PanzerKunst:
Between:
A. Halve the plot filter every couple years
B. Increase k value every couple years

As far as I understand, B will require many farmers to replot to k+1 every couple years, whereas A doesn't require such a periodic replot. If that is indeed the case, I think this CHIP is a great proposal, and I support it.

CryptoBlockchainTechnologies:
This is a good proposal to ensure there is a constant cadence in hardware requirements, however I think the two year implementation schedule is a bit aggressive. Why was two years chosen, was it arbitrary, or based on ASIC game theory?

I think it makes more sense to tie in the halving of the plot filter with the halving of coin emissions every three years to prevent them from occurring too close together. More specifically six months prior to the coin halving we would get a plot filter halving. This would allow farmers to upgrade their hardware prior to coin halving in what is typically a positive market environment. We could also do it six months after coin halving but I think prior to coin halving makes the most sense. regardless arbitrarily picking two years does not seem like a good idea and is a tad extreme.

So if we decide three years makes more sense and this CHIP passes, the first filter halving would be 10/1/2023, six months before coin emission halving on 4/1/2024. That is approximately eight months from today.

mech2you:
To me it sounds like just pushing the problem into the future. What will happen if in a few years the network size is 2 times bigger and the nodes are mostly at k32? There is no incentive to plot on k33, k34 or kXX. However, reducing the size of the filter does not reduce the memory requirements for creating a plot. At some point, ASIC miners will come and replace the GPUs.  Therefore, in my opinion, increasing the K size oder changing the plot format makes more sense, even if not everyone is willing to do it. By increasing the memory requirements when creating the plots, this ensures that no ASICs are developed in this direction because RAM memory is still expensive. Better to lose some network size now than face the same problem in 7 years. I myself have a 1.5PiB farm with about 70% k32 plots. Every farmer knew that the moment will come when the K32 size is not enough ;)

Motophan:
Here is how I see it -> lets say chia corp goes bankrupt. I do not see considerations for this scenario. 

Startup tech companies have a very high likelihood (>90%) of going bankrupt. Ones who have less revenue than expenses are more likely to fail. 


This CHIP will send us to a plot filter of 32 and stay there forever. 

I suggest doing a plot filter change on a as needed basis and not schedule them as its (extremely) likely chia corp will not be around to reset the plot filter and set it to k33. Chia corp will not have to live w/ the problems they create. We will.

reythia:
Reducing plot filter towards 32 seems sensible, and preferable as the first lever, but is difficult to consider in isolation.

At some point a lower filter necessitates an increase in k.

Cumulative 16x decrease in farmable space with a given setup will catch out many farmers, especially those about to plot with compression levels that appear to have plenty of headroom. They'll either have to upgrade hardware with increasing aggression or replot in some form - to higher k or lower compression.

If k=34 by the time filter=32 that reduces the increase in farming demands to 4x over ~8 years. Feels like a sensible balance on multiple fronts and provides a predictable enviornment for farmers.

Perhaps a cadence of filter-filter-k-filter-filter-k? With an 18 month or "half halving" interval.

SlowestTimelord:
> I suggest doing a plot filter change on a as needed basis and not schedule them as its (extremely) likely chia corp will not be around to reset the plot filter and set it to k33. Chia corp will not have to live w/ the problems they create. We will.

The chain does not _need_ Chia Network Inc to incorporate any changes. The client can be forked and the farmers can decide which version to run if CNI disappeared off the face of the earth tomorrow. Also the scheduled plot filter change is purposely more aggressive than needed because relaxing the plot filter schedule would be a soft fork while accelerating it would be a hard fork -- so you'd want to default to a more aggressive schedule to be conservative. It would be far more disruptive to make plot filter changes "as needed" since each would need to be a hard fork.

danieljperry:
> Thanks for writing such a thorough explanation of the problem and proposed solution @jmhands! A few clarifying questions that I have already asked you in other channels, but adding here for completeness:
> 
>     1. Does reducing the plot filter by half also double plot "decompression" farming workload? Or is it not exactly proportional?
> 
>     2. Upon reaching size 32 plot filter, does that imply increasing minimum k-size to k33 is the logical next proactive step? At that time would plot filter reset to 512 or stay at 32?
> 
>     3. Is increasing the minimum k-size a soft fork?
> 
>     4. In the event of a hard fork, would that be an opportunity to introduce other pending changes that require a hard fork? What such changes might those be?
> 
>     5. Would increasing the number of signage points per subslot (to decrease the time between infusion points) be a viable option as well or is that too drastic a change to the consensus?

1. Yes, all other things being equal, this would double the farming workload. However, keep in mind that the workload is minuscule to begin with.
2. We have no plans for changing either the filter or the minimum k size after this proposal runs out. The reasons for this are that 
  a. At a plot filter of 32 (keeping in mind that this will be 8 years from now), low end machines will still be able to farm Chia using HDD storage, and 
  b. It's difficult to predict the state of the storage industry that far into the future. Conventional wisdom says that SSD storage will be cheaper than HDD storage at that time, which will mean that most of the netspace will be on SSDs as well. If that's the case, there may not be a need for a filter at all, but we simply don't know if this prediction will come true yet.
3. Yes. A soft fork is a tightening of the rules, and a hard fork is a loosening of the rules.
  * Increasing k would mean that some plots (k-32) would no longer work, therefore the rules would have been tightened so it would be a soft fork.
  * Reducing the filter would mean that some plots that did not pass the filter before the change would pass the filter after the change. This would be a relaxation of the rules and therefore would be a hard fork. Following this logic, if we were ever to increase the filter, it would be a soft fork.
4. Potentially yes, but we don't have any such plans at this time. If such plans did emerge, they would be proposed in a new CHIP. This is because each CHIP needs to focus on a single proposal.
5. To my knowledge changing the number of signage points per subslot wasn't discussed, though other consensus changes were. My initial thought about this idea is that it would make plot grinding more expensive and more difficult, but it wouldn't achieve significantly more than simply decreasing the filter. This would also be a large consensus change, it could affect the stability of the network, and it could potentially open the network to new attack vectors. While alternate proposals are certainly welcome, I don't see this one being viable (unless I'm mistaken about something fundamental).

danieljperry:
> A few more thoughts to stir the pot in favor of Option 2 (k-size change):
> 
>     * If there were ever a time to declare timelines for a minimum k-size, it's probably now while folks are considering replotting anyway.
> 
>     * I don't think a proposal on plot filter decrease is complete without considering when/how minimum k-size requirement might fit into the ~10 year timeline.
> 
>     * Plot filter changes + decompression considerations adds a dimension that many farmers may have trouble grasping or considering whereas a minimum k-size timeline is easier to grasp (say, k33 phases in 2025, k34 phases in 2028 etc.)
> 
>     * Given the unpredictability of future hardware and plotting/farming advances, there's a chance another hard fork to change the timeline of plot filter changes may be necessary whereas as Bram has mentioned in the past a minimum a k-size change would be a soft fork (I'd like to confirm whether or not this is the case).
> 
> 
> > Increased hardware requirements for plotting. A `k=34` plot would be very difficult to create in RAM without using a high-end server
> 
> Maybe it's my experience from early plotting days but I've always viewed the ability to plot all in RAM (and now to plot with GPU+RAM) as a nice-to-have and not something farmers should feel is necessary or feel entitled to be able to do.

These are all fair points and the community will need to carefully weigh the options, just as CNI already has done.

danieljperry:
> What is the plan after 2030?

No plans yet. We have a decent idea of what the storage and GPU industries will look like in a few years, but it's difficult to predict what will happen 7+ years from now. However, the two primary mitigations we have against plot grinding are -- and likely will remain -- reducing the filter and increasing the minimum k.

danieljperry:
> passes, the first filter halving would be 10/1/2023, six months before coin emission halving on 4/1/2024, which is approximately eig

JM is the expert here, but I can say that the timeline from this proposal was not chosen arbitrarily. It was chosen based on the likely advancements in the GPU, memory, and storage industries over the next few years. The exact block heights where the changes will be implemented is something that can be discussed on Discourse, which will be set up soon.

danieljperry:
> To me it sounds like just pushing the problem into the future. 
Correct, this proposal is meant to make plot grinding uneconomical over the next 8 years, but not forever.

> What will happen if in a few years the network size is 2 times bigger and the nodes are mostly at k32? 
The network size has nothing to do with this proposal. Most plots will likely still be k-32 because they are the easiest to create, though larger plots do require less resources while farming.

> At some point, ASIC miners will come and replace the GPUs.
We don't foresee this happening (assuming you mean "plotters" and not "miners"). If you have any data to back this up, please share it.

lord-icon:
When considering K32 on K33:
Is it technically possible to convert 2xK32 plots to a K33?
Replotting 10,000 K33s fails the CHIA policy.

danieljperry:
It might be possible but it will require more time and energy than simply creating a new k33. However, this proposal does not require replotting. That decision remains entirely yours, regardless of whether the proposal succeeds or fails to gain adoption.

RudolfAchter:
This is the Plot Filter proposal from [GPU Plotting is Real – and Very Fast](https://www.chia.net/2023/01/20/gpu-plotting-is-real-and-very-fast/) right? 
Am i reading correct if this plot filter proposal is implemented then there will be a hard fork in the blockchain?
So farmers have to update to a specific software version before a certain block is reached?
Would have no problem with it. Just want to clarify.

I think also with blockchain there is always a need to upgrade software to adapt to technology development.

danieljperry:
> This is the Plot Filter proposal from [GPU Plotting is Real – and Very Fast](https://www.chia.net/2023/01/20/gpu-plotting-is-real-and-very-fast/) right? Am i reading correct if this plot filter proposal is implemented then there will be a hard fork in the blockchain? So farmers have to update to a specific software version before a certain block is reached? Would have no problem with it. Just want to clarify.
> 
> I think also with blockchain there is always a need to upgrade software to adapt to technology development.

This is correct, yes.

keliew:
>    a. At a plot filter of 32 (keeping in mind that this will be 8 years from now), low end machines will still be able to farm Chia using HDD storage, and
Is the definition of a "low end machines" still a RPi4? Or will the minimum requirement increased due to this change. Obviously in this scenario, the farmer is not farming compressed plots.

danieljperry:
This CHIP is now a `Draft`. If it is accepted, it will require a hard fork of Chia's blockchain. Therefore it is imperative that we have consensus from the community. Please continue to leave your reviews here.

You can also discuss this CHIP in the #chips channel on Keybase, or on Discourse:
https://developers.chia.net/t/chip-0012-plot-filter-reduction/810

Additionally, we will hold a public Zoom call to discuss this CHIP at 8 AM PST on Wednesday, February 22. See the #chips channel on our Keybase for details.
https://keybase.io/team/chia_network.public

danieljperry:
> > a. At a plot filter of 32 (keeping in mind that this will be 8 years from now), low end machines will still be able to farm Chia using HDD storage, and
> 
> Is the definition of a "low end machines" still a RPi4? Or will the minimum requirement increase due to this change. Obviously in this scenario, the farmer is not farming compressed plots.

The definition of "low end machine" will likely remain within the same genre for the foreseeable future. For example, the min spec machine could become a RPi5, but it won't be a workstation.

danieljperry:
Here is a recording of the discussion of this CHIP:
https://youtu.be/hf01z8Uhj1Y

Please continue to leave your reviews here.

SlowestTimelord:
Thanks for the informative public zoom call today. Wanted to summarize some thoughts:

- Option 1 (plot filter reduction) makes much more sense to me than Option 2 (increase minimum k-size) if only because Option 2 likely *still* requires a plot filter reduction, which almost makes Option 1 strictly better. But I would still like to see documented guidance about what situations *would* warrant a k-size increase? If I understand correctly from the call today, it is when multi-GPU plotting can economically do k32s entirely in combined VRAM.

- There needs to be more definitive direction/education for farmers looking to do compressed plot farming at higher compression levels (C6+) because plot filter reduction will proportionally increase decompression workload and hence proportionally reduce maximum farm size (by up to 16x). Plotting and farming now requires longer term planning for farmers looking to min/max their compression.

- Probably a moot point but re: validity of proofs from older clients. Plots that pass the 512 filter will also pass lower plot filters. It was brought up on the call that clients that don't upgrade have a high likelihood of farming on the wrong chain due to rejecting valid proofs that the most of the network sees as valid.
Seeing "[Farm Sooner](https://roadmap.chia.net/c/54-farm-sooner?utm_medium=social&utm_source=portal_share)" on the 2023 roadmap is interesting because I think a "light wallet" farmer relying on full node peers might be able to continue farming 512 filter plots at their own disadvantage. This is only relevant if "Farm Sooner" is released well before a plot filter reduction.

- Given the expected eventual SSD flippening, wouldn't it make sense to also put in a plot filter reduction to 1 in the distant future to reduce likelihood of another hard fork? e.g. plot filter to 1 reduction in 2040. This may be aggressive but again, this can be pushed out via soft fork if we realize HDD farming is still a non-negligible portion of the netspace in 2040.

danieljperry:
Thanks for these thoughtful comments. We will take them all into consideration.

JustAlias:
I have 100,000     k32   plots😓😓

danieljperry:
> I have 100,000 k32 plots😓😓

This CHIP does not affect the validity of k32 plots.

babipanghang:
So, from what i understand from a recent blogpost:

> Thankfully it requires a very large cluster of GPUs and a tremendous amount of power to perform something like this today, and it is not economical even with the extended leverage.

In this context, sure one is not likely to invest in hardware for plot grinding like this, but what about people who already own a lot of GPUs? Chinese mining farms? People who have a bunch of hardware laying around after ethereums recent move to POS or otherwise? Will it be economical to them and if so, will lowering the plot filter be likely to deter them?

lord-icon:
GPU mining should generally be prevented. Then you have to drastically reduce the plot reporting time.

Reason: CHIA is still dependent on investors.
CHIA advertises with GREEN-IT. If CHIA becomes POW (even partially) then it will be detrimental to investors.

hoffmang9:
> So, from what i understand from a recent blogpost:
> 
> 
> 
> > Thankfully it requires a very large cluster of GPUs and a tremendous amount of power to perform something like this today, and it is not economical even with the extended leverage.
> 
> 
> 
> In this context, sure one is not likely to invest in hardware for plot grinding like this, but what about people who already own a lot of GPUs? Chinese mining farms? People who have a bunch of hardware laying around after ethereums recent move to POS or otherwise? Will it be economical to them and if so, will lowering the plot filter be likely to deter them?

In the next 4 years without this change, the only economically viable attack would be a short term rental attack. Such an attack would use a material percentage of all GPUs that exist. The plot filter change cuts the viability of any such attack in half (doubles the required amount of GPUs to attempt it.) As to farming over weeks or months, using GPUs instead of just using them to plot and fill space remains un economic for a while without this change and with this change, doubles the cost for good measure.

danieljperry:
> GPU mining should generally be prevented.

If an individual were able to construct a GPU farmer, they could keep it for themselves or rent/sell its usage to gain an economical advantage over everyone else. Our approach therefore is to create the best tech possible and keep it open sourced. That way, no individual has an advantage over everyone else.

That said, we don't believe there is a viable way to prevent GPU farming altogether. CNI certainly has no ability to censor anyone from using the Chia blockchain. If you do have an idea to prevent GPU farming from occurring (along with the technical details to implement your idea), you are welcome to submit a CHIP. CNI, along with the entire Chia community, will take your idea into consideration.

SlowestTimelord:
> That said, we don't believe there is a viable way to prevent GPU farming altogether.

I believe the comment was in support of the plot filter reduction as a means to reduce the economics of plot grinding specifically with GPUs.

CryptoBlockchainTechnologies:
@lord-icon You may not understand why over 24 years ago Nvidia created the term.GPU and moved the current graphics chip from motherboards and made a discrete graphics card. This is because unlike CPUs, GPU are made for one thing, compute at extremely efficient and powerful levels.

I don't blame you though, I blame CNI for their campaign against GPUs and not putting out real numbers like I have showing the community how much more efficient GPU farming is. Yes this CHIP will be used to hold at bay GPU farming even though it is more efficient than buying more harddrives because they can't build their own GPU farmer.

My 5PB farm uses only 190 watts more to run C7 plots on a 3060 GPU. I gain 37% with an additional effective space of 1.85PB. This would be equvalent to adding 103x18TB enterprise drives. Since each drive uses 9W this would add another 927W to my farm instead of 190W for compressed plots using a GPU.

SlowestTimelord:
> I don't blame you though, I blame CNI for their campaign against GPUs and not putting out real numbers like I have showing the community how much more efficient GPU farming is. Yes this CHIP will be used to hold at bay GPU farming even though it is more efficient than buying more hard drives and farming with a CPU. Either they can't or don't want to build their own GPU farmer possibly because of their ties to the hard drive industry.

To be clear, this CHIP is not to prevent GPU Farming. This is to prevent the possibility of GPU _grinding_. As far as I know GPU plotting + farming is on the roadmap and in fact encouraged for farmers (such as yourself) where the efficiency makes sense.

The plot filter reduction *will*, as a side effect, reduce efficiency of GPU farming however, but that is not its goal or purpose.

hoffmang9:
> @lord-icon You may not understand why over 24 years ago Nvidia created the term GPU and moved the current graphics chip from motherboards and made a discrete graphics card. This is because unlike CPUs, GPU are made for one thing, compute at extremely efficient and powerful levels.
> 
> My 5PB farm uses only 190 watts more to run C7 plots on a 3060 GPU. I gain 37% with an additional effective space of 1.85PB. This would be equivalent to adding 103x18TB enterprise drives. Since each drive uses 9W this would add another 927W to my farm instead of 190W for compressed plots using a GPU.
> 
> I don't blame you though, I blame CNI for their campaign against GPUs and not putting out real numbers like I have showing the community how much more efficient GPU farming is. Yes this CHIP will be used to hold at bay GPU farming even though it is more efficient than buying more hard drives and farming with a CPU. Either they can't or don't want to build their own GPU farmer possibly because of their ties to the hard drive industry.

You are conflating the wrong things. This CHIP addresses the problem of people substituting only GPU grinding for storing plots on disk. We totally agree that plotting with a GPU is more efficient and you factually misstate our position on that as GPU plotting and farming are almost shipped to the beta program. This is about keeping GPU farms from simulating space in a proof of work style.

CryptoBlockchainTechnologies:
> You are conflating the wrong things. This CHIP addresses the problem of people substituting only GPU grinding for storing plots on disk. We totally agree that plotting with a GPU is more efficient and you factually misstate our position on that as GPU plotting and farming are almost shipped to the beta program. This is about keeping GPU farms from simulating space in a proof of work style.

Then explain why you are proposing to halve the plot filter three times in a little more than 4 years when you just said the above statement

> In the next 4 years without this change, the only economically viable attack would be a short term rental attack.

2024-256
2026-128
2028-64

SlowestTimelord:
> Then explain why you are proposing to halve the plot filter three times in a little more than 4 years when you just said the above statement

It's all explained in the CHIP -- which part do you specifically have a question about?

lord-icon:
is that data you are already running?  Or is that the extrapolation according to the Excel list?  Because hardly anyone achieves this.  I also doubt that you can manage 5PB with a 3060 (without ti?).  or: final 6.85PB.  But yes.  I don't really question its effectiveness either.  Much more the procedure, which is increasingly moving in the direction of POW.  And CHIA was never "born" for that.  What is the next level?  ASIC?  C30?

SlowestTimelord:
> is that data you are already running? Or is that the extrapolation according to the Excel list? Because hardly anyone achieves this. I also doubt that you can manage 5PB with a 3060 (without ti?). or: final 6.85PB. But yes. I don't really question its effectiveness either. Much more the procedure, which is increasingly moving in the direction of POW. And CHIA was never "born" for that. What is the next level? ASIC? C30?

I implore you to read [Chia.net's blog posts detailing plot compression and plot grinding](https://www.chia.net/2023/01/20/plot-compression-is-here/). In particular understanding the exponential increasing workload of "decompression" to see why there is no "next level".

CryptoBlockchainTechnologies:
> is that data you are already running? Or is that the extrapolation according to the Excel list? Because hardly anyone achieves this. I also doubt that you can manage 5PB with a 3060 (without ti?). or: final 6.85PB. But yes. I don't really question its effectiveness either. Much more the procedure, which is increasingly moving in the direction of POW. And CHIA was never "born" for that. What is the next level? ASIC? C30?

Yes this is real time farm data

lord-icon:
> Yes this is real time farm data


hmmm... I'm looking again for the 2 reddit threads that achieve completely different (much higher) values.  Also with the graphics card.  But I have to look it up again to be sure.  But I think it was also a 3060. And it was much more likely to be fully utilized.  (no 6.85PB).  May I ask which one you have exactly?  Manufacturer and GB.  6GB or 12GB?  ti or without.  Then I can check again whether your setup is better.  Thanks

SlowestTimelord:
> Yes this is real time farm data, in theory my max farm size WITH a 512 plot filter is 8.4PB. I have plans to expand to 10PB this fall and move to a 3060Ti. However in 2024 if this passes my max farm size will be cut in half to 4PB. This CHIP is way to aggressive especially when even Gene is admitting it will not be cost effective for the next 4 years. I would much rather prefer 3 years just before every halving. Two years on a ghost you are afraid of is ridiculous and will only keep most people out of GPU farming.

Again, Gene is talking about cost effectiveness of *grinding* not *farming*.

Thinking the plot filter reduction schedule is too aggressive is a valid point so thank you for your suggestion of 3 years instead of 2. Do keep in mind making the schedule less aggressive later (as we see how technology progresses and in particular the PCIe 5.0 adoption) can be done with a soft fork but it would be very disruptive to realize 3 years is not aggressive enough later and hard fork again to move it to 2 years.

> hmmm... I'm looking again for the 2 reddit threads that achieve completely different (much higher) values. Also with the graphics card. But I have to look it up again to be sure. But I think it was also a 3060. And it was much more likely to be fully utilized. (no 6.85PB). May I ask which one you have exactly? Manufacturer and GB. 6GB or 12GB? ti or without. Then I can check again whether your setup is better. Thanks

We should probably keep the discussion in this thread focused on the CHIP and leave the farming optimization discussions to a separate conversation.

CryptoBlockchainTechnologies:
> > Yes this is real time farm data
> 
> hmmm... I'm looking again for the 2 reddit threads that achieve completely different (much higher) values. Also with the graphics card. But I have to look it up again to be sure. But I think it was also a 3060. And it was much more likely to be fully utilized. (no 6.85PB). May I ask which one you have exactly? Manufacturer and GB. 6GB or 12GB? ti or without. Then I can check again whether your setup is better. Thanks

While I prefer EVGA GPUs (I use them for plotting) I have an ASUS TUF GAMING GeForce RTX 3060 12GB. These are cards I used for mining Ethereum that I kept when I sold my farms. I kept the EVGA because they no longer make them and they always made the best GPUs. I farmed with their cards for 5 years 24x7 and never had any fans fail. My plan as the plot filters get cut in half is to use my mining motherboards where I can use up to 13XGPUs on a single rig to farm with. Ironically the GPUs don't need as much PCIe bandwidth when farming.

CryptoBlockchainTechnologies:
> Again, Gene is talking about cost effectiveness of _grinding_ not _farming_.
I am ignore  you because either you are a bot or someone who has only canned replies like when you call IT for support. You need to rethink the discussion and how it affects GPU farmers.

> Thinking the plot filter reduction schedule is too aggressive is a valid point so thank you for your suggestion of 3 years instead of 2. Do keep in mind making the schedule less aggressive later (as we see how technology progresses and in particular the PCIe 5.0 adoption) can be done with a soft fork but it would be very disruptive to realize 3 years is not aggressive enough later and hard fork again to move it to 2 years.
If you were paying attention or are a bot you would have known I made this suggestion when it was CHIP-53. So going back to ignoring your rants.

hoffmang9:
Reducing the plot filter will actually have the side effect of making more farmers switch from CPU farming to GPU farming. I’m concerned that the known negatives of running really big farms are clouding some folks judgement.

CryptoBlockchainTechnologies:
> Reducing the plot filter will actually have the side effect of making more farmers switch from CPU farming to GPU farming. I’m concerned that the known negatives of running really big farms are clouding some folks judgement.

I have been seeing quite the opposite. Small farmers would like to replot using the highest CPU compression possible but have say a 200TB farm using an old CPU and C5 compression puts them close to hitting max farm size but makes the most sense for them to do all the replotting. Fast forward only a year to the first plot filter halving and they are now below what their system is capable of farming. Most are deciding to not replot because it is simply not worth it when considering the aggressive plot filter halvings. But this is what you guys want apparently.

I have no issues keeping up and I am sure a lot of the other large farmers will be able to as well. Just understand who will still be around in a few years as we push down profitability of those who don't use compression. This aggressive plot filter halving will have huge implications on your farming community, mostly the small ones. I don't come here to campaign for myself, I will be just fine. Rather I prefer a strong and thriving community and we need to ensure a change like this is really needed.

jmhands:
If you watched the call we discussed 3 year cadence being reasonable when also using the filter change in CHIP13. Understand this will impact farming strategy and we are developing tools to help farmers of all sizes simulate compute overhead and estimated max farm size for now and the future

CryptoBlockchainTechnologies:
> If you watched the call we discussed 3 year cadence being reasonable when also using the filter change in CHIP13. Understand this will impact farming strategy and we are developing tools to help farmers of all sizes simulate compute overhead and estimated max farm size for now and the future

Yes I saw that and was encouraged, believe it is a combination of CHIP-13 also. It is a great idea and I think makes more sense.

SlowestTimelord:
Great points @Voodoo-Chia. There's also a high likelihood SSDs will make up much more of the netspace by the 3rd or 4th plot filter reduction, which also dampens the impact of look ups (though not decompression).

Now that we have established block height (date) proposed, I would like to see this proposal move forward in the CHIP process along with #57. I have not seen any objections to the CHIP that haven't already been addressed by the comments above.

CryptoBlockchainTechnologies:
@SlowestTimelord 
I will 2nd that motion. My original objection I stated here was to extend plot filter having to two years. Now that this has been done, it has my support.

danieljperry:
Thanks everyone! Given your your feedback, along with projections for the advancement of hardware, we have updated the proposal to use a 3-year cadence for reduction, starting in one year. We will be moving this CHIP along in the review process soon. To make it absolutely clear, here is the current proposed schedule:

| Block height | Month/Year (approx) | Filter size |
|------------: | ------------------: | ----------: |
|  `5 496 000` | June 2024           | 256         |
| `10 542 000` | June 2027           | 128         |
| `15 592 000` | June 2030           |  64         |
| `20 643 000` | June 2033           |  32         |

Additional feedback and reviews are still welcome!

danieljperry:
This CHIP is in `Review`. Please make your final reviews soon.

digitalspaceport:
Full support, like the updated filter reduction cadence. Glad to see it in review with specific blocks. Could we get estimated dates for those blocks?
________________________________

SlowestTimelord:
The month estimates (especially for later reductions) might be as accurate as we can get. With the ASIC timelords coming online the exact corresponding date could have multiple days/weeks of variation?

danieljperry:
We will minimize the impact of the ASIC timelords by introducing them slowly. However, other events such as a rapid increase in netspace could move the block heights by a few hours or days. We won't be able to give a more granular estimate than month and year.

geraldneale:
Let's call this what it is, a tendency away from PoST and towards PoW.

SlowestTimelord:
> Let's call this what it is, a tendency away from PoST and towards PoW.

This is exactly the opposite. It's a lever to ensure PoST doesn't become PoW.

danieljperry:
> Let's call this what it is, a tendency away from PoST and towards PoW.

Can you explain in more detail why you hold this opinion?

geraldneale:
PoST always had a POW component therein because there is no other way to make plots other than doing work. For example, doubling lookups requires doubling the work for this component as this CHIP clearly describes. Also, replotting means more work. Also, the new compression algorithms move work to the CPU compared to the old. 

PoST advocates, as I still consider myself to be, should always recognize these facts. Proof of Space Time would have been better labelled Proof of Space Time _and Work_(PoSTW) from the beginning because work is NOT removed by Chia Consensus only being managed better, but the key to managing is recognizing it imo.
 
There is clearly a precedent in Bitcoin that POW works well, but over-reliance on work tends to centralize and Chia is tending in that direction with this CHIP and other updates imo. I'm not offering an alternative, because perhaps it is inevitable, except enhanced disclosure so more folks can understand why it is necessary. I think this is important for the sake of mass uptake and therefore security.

danieljperry:
Thanks for the feedback!

Responses in line:
> PoST always had a POW component therein because there is no other way to make plots other than doing work.

True, but irrelevant for this CHIP.

> doubling lookups requires doubling the work for this component as this CHIP clearly describes.

Also true, but it only doubles the lookups a farmer must perform. No new hashes are generated. No new data is written. It's doubling the amount of work, but reading data from a disk is not the same as the Proof of Work consensus.

> the new compression algorithms move work to the CPU compared to the old.

Irrelevant -- This CHIP is reacting to the new compression algorithms, but it is not the cause of those algorithms, which would exist regardless of the existence of this CHIP.
> replotting means more work.

Irrelevant -- This CHIP does not necessitate any replotting.
 
> There is clearly a precedent in Bitcoin that POW works well, but over-reliance on work tends to centralize and Chia is tending in that direction with this CHIP and other updates imo. 

Do you have other reasons to believe that Chia is trending in the direction of PoW that are specific to this CHIP?

> [...] enhanced disclosure so more folks can understand why at least a minimum amount of work is necessary, but even with incremental increases it is not necessary in the same abundance as Bitcoin. I think this is important for the sake of mass uptake and therefore security.

Agreed, full disclosure is very important here. That's why this CHIP has been open for five months and counting.

hoffmang9:
> PoST always had a POW component therein because there is no other way to make plots other than doing work. Also, doubling lookups requires doubling the work for this component as this CHIP clearly describes. Also, the new compression algorithms move work to the CPU compared to the old. Also, replotting means more work.

Speaking holistically and not to just this CHIP - on net the cost of plotting in both time and CPU power has fallen/is being amortized over farming. When looking at a new entrant, the amount of time and CPU/GPU that they have to allocate to be up and providing material additional security is greatly decreased by all these changes.

As Chia blockchain scales, that next 25 EiB will be brought on much more cheaply than before the new methods to create smaller plots on disk existed.

geraldneale:
If there's one thing I "know" it is that nobody knows the future. Where does it say in this CHIP why the filter adjustment can't be more like Bitcoin where the difficulty adjustment is based on feedback from the network, not the solely the calendar(e.g. one year out and then every 3 years)? Sorry if I missed it. Since it requires a hard fork, perhaps this is an opportunity to build something that presumes less about the future.

hoffmang9:
The designed from the start mechanism to defend against the ongoing increase in memory bus speed between parallel processors is the raising of the minimum K size. This was an opportunity to use a less blunt instrument, but if the plot filter decreases are not enough, then we revert to the original plan that's always been there.

geraldneale:
> * ass the 512 filter will also pass lower plot filters. It was brought up on the call that clients that don't upgrade have a high likelihood of farming on the wrong chain due to rejecting valid proofs that the most of the network sees as valid.

I was wondering why this wasn't programmed in from the start since it seems so systematic, but I used up all of my adversarial stance question points today ;) Thank you for answering the implied question.

danieljperry:

danieljperry:

Rigidity (approved):

PR #81, opened 2023-08-21
CHIP-0020: Wallet Hinted Coin Discovery

Rigidity:


danieljperry:
This CHIP presents an optional method for Chia wallets to implement hinting. It is being opened directly to a status of `Review (Fast Track)` because it is already in use in Chia's reference wallet.

Please leave your reviews as comments in this PR.

danieljperry:
We will be holding a public Zoom call to discuss this CHIP on September 5 at 1500 UTC (8 AM PDT, 11 PM China). See the #chips channel in our [Discord](https://discord.com/invite/chia) for more info.

Rigidity:
In addition to rewording the CHIP a bit to make it easier to follow, some questions brought up on the call have been addressed in the CHIP. A review would be appreciated.

**Is the zero-index of the memo list always interpreted as a hint?**
Only if the first memo is 32 bytes long.

**If you want to include a memo on an unhinted coin, do you need to set the hint to null?**
This CHIP's wording previously implied that the hint is separate from the other memos in the list. This is not the case, and has been corrected. As such, even if a memo is 32 bytes, it does not need to be in a separate position than a hint. It would just be considered a "hint" to that 32 bytes in _addition_ to a regular memo for other purposes (and assuming those 32 bytes aren't a valid puzzle hash that a wallet would look up, it would have no effect).

**Is it possible to DoS a wallet or node by including a large hint?**
No, since a hint must be exactly 32 bytes. If the first memo is larger than this, it would be discarded and not stored in the database. You also still have to pay the regular CLVM cost of 12,000 per byte for memos.

**Are RPCs in the scope of this CHIP?**
Documentation of RPCs and implementation details are not in the scope of this CHIP, however a link has been added to the `get_coin_records_by_hint` RPC.

**What is the CLVM cost of a hint?**
A hint is 32 bytes, and each byte has a cost of 12,000, so the total cost of the hint is 384,000.

danieljperry:
All questions have been answered regarding this CHIP, and it appears to have achieved community consensus. It is now in `Last Call`. If no further changes are required in the next two weeks, this CHIP will be moved to `Final`, after which changes (other than errata) will no longer be allowed.

danieljperry:

danieljperry (approved):

PR #90, opened 2023-11-26
init the chip 0023: on-chain vote

wizicer:
I raised the CHIP while some remaining works exist:

- [ ] Implement the vote function UI in Pawket.
- [ ] Implement the test cases.
- [ ] Try run with the community.

### DIFF
--- a/CHIPs/chip-0023.md
+++ b/CHIPs/chip-0023.md
@@ -1,0 +1,152 @@
+CHIP Number   | 0023
+:-------------|:----
+Title         | On-chain Vote
+Description   | A standard for implementing simple on-chain vote on Chia's blockchain
+Author        | [Shuang Liang](https://github.com/wizicer) ([@icerdesign] on Twitter)
+Editor        | 
+Comments-URI  | [https://github.com/Chia-Network/chips/pull/90](https://github.com/Chia-Network/chips/pull/90)
+Status        | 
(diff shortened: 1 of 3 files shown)


Yakuhito:
Hey there,

Two points I see as a community dev:
 - This proposal doesn't contain a mechanism to count votes on-chain, which would be required by DAOs
 - Chia has been working on a DAO primitive for some time now, and the puzzles are already on the 'main' branch - having two voting/DAO standards would be confusing

wizicer:
> Hey there,
> 
> Two points I see as a community dev:
> 
> * This proposal doesn't contain a mechanism to count votes on-chain, which would be required by DAOs
> * Chia has been working on a DAO primitive for some time now, and the puzzles are already on the 'main' branch - having two voting/DAO standards would be confusing
> 
> That being said, I'm super happy to see the Pawket team again!

Good to know CNI is working on the primitives, hope I can use them before I implement this.
(maybe we can at somebody to let them know our keen anticipation to this feature.)

For the sake of simplicity, on-chain count is intended omitted. By the nature of transparent of every vote is on-chain, achieving consensus via off-chain count becomes straightforward.

danieljperry:
This CHIP is now `Stagnant` due to having been inactive for over six months. Please comment here if you would like this CHIP to be reopened as a `Draft`.

PR #95, opened 2023-12-27
Long term grinding protection - first draft

SlowestTimelord:
Proposal for plot filter removal to address long term grinding protection.

### DIFF
--- a/CHIPs/chip-long-term-grinding-protection.md
+++ b/CHIPs/chip-long-term-grinding-protection.md
@@ -1,0 +1,107 @@
+CHIP Number   | ????
+:-------------|:----
+Title         | Long term grinding protection
+Description   | A hard fork to aggressively reduce grinding economics, can be adjusted with soft forks later.
+Author        | [Slowest Timelord](https://github.com/SlowestTimelord)
+Editor        | < Creator must leave this blank. Editor will be assigned.>
+Comments-URI  | < Creator must leave this blank. Editor will assign a URI.>
+Status        | < Creator must leave this blank. Editor will assign a status.>
+Category      | Standards Track
+Sub-Category  | Core
+Created       | 2023-12-27
+Requires      | None
+Replaces      | None
+Superseded-By | None
+
+## Abstract
+In the design of Chia's consensus, there are several built-in levers that can be adjusted to reduce the cost-effectiveness of plot grinding. These options were assessed in CHIP-0012 and ultimately, four steps of plot filter reduction over 10 years (down to 32 in 2033) was accepted and implemented with a hard fork. Hard forks are not desirable and more will be necessary in the future to keep up with technological advances. This CHIP proposes to implement a "final hard fork" *now* with **intentionally aggressive timelines** towards eventual plot filter removal with the understanding that if deemed too aggressive, timelines can be pushed back later with a soft fork.
+
+## Motivation
+The primary motivation to reduce the cost effectiveness of plot grinding was [described well in CHIP-0012](https://github.com/Chia-Network/chips/blob/main/CHIPs/chip-0012.md#motivation), the reasoning here remains the same. Additional motivation for outlining a more aggressive and longer term plan against plot grinding include:
+- Recognizing recent developments in GPU plotting efficiency and hence grinding effectiveness.
+- Recognizing a potential ~3x gain in plot grinding effectiveness that may not have been evident at the time of CHIP-0012 discussion. (A moot point if CHIP-0013 is implemented.)
+- A desire to not have further hard forks every 3-5 years to keep up with plot grinding effectiveness. Instead have one "final" aggressive hard fork now that can be adjusted as needed with soft forks.
+- Provide transparency to farmers as to what the future of Chia farming looks like.
+- Recognize an eventual flipping of SSD vs HDD cost effectiveness and future farming considerations for when majority of netspace will be on flash storage.
+
+It is important to acknowledge that this CHIP is *not* motivated by:
+- The [increasing share of netspace using plot compression](https://xch.farm/compressed-netspace/)
+- Increased centralization caused by some farming pools not following the official pooling protocol
+- Any increased energy use that comes from the above
+
+However, by reducing plot grinding effectiveness, there *will* be a side effect of impacting farmers and third parties that utilize or otherwise benefit from compressed plot farming, potentially requiring them to pull ahead the rethinking of their long term farming strategy. There will also be long term impact to farmers utilizing slow large capacity drives. See the sections below on impact for more details.
+
+### Proposal
+
+This CHIP is a proposal to aggressively pull ahead the plot reduction schedule beyond the 256 reduction to every 1 year instead of every 3 years. It also introduces new timelines for further reductions all the way to plot filter removal (Filter size = 1), at which point it can be expected for much of the netspace to be on SSDs where [proof lookups aren't bottlenecked by disk latency](https://chiapower.org/Power/powerssd).
+
+| Block height | Month/Year (approx) | Filter size | vs CHIP-0012           |
+|------------: | ------------------: | ----------: | ---------------------: |
+|  `5 496 000` | June 2024           | 256         | Unchanged              |
+| `10 542 000` | June 2025           | 128         | June 2027 -> June 2025 |
+| `15 588 000` | June 2026           |  64         | June 2030 -> June 2026 |
+| `20 634 000` | June 2027           |  32         | June 2033 -> June 2027 |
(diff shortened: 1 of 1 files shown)


digitalspaceport:
I can see a change in the plot filter reduction schedule as a specific type of mitigation, however I think this is ultimately the wrong path. I would like to voice that I DO feel a change is needed from what we are seeing happen in the compressed ecosystem to address the composed balance of plotted netspace, rewards, and most importantly the security of the chain. 

The levelers that at the time were available to be pulled to provide ongoing security of the chain for CHIP 12 addressed what at the time seemed like a specific case around GH. We have a very different competing product to GH out now and indeed its impacts on NC have been dramatic for Chia. I think reassessing the levers pulled from a K minimum size need to be reevaluated. To be clear I am suggesting a change to the minimum K size is something that should be on the table. There are a wide variety of benefits that approach could have as well as a seperate set of challenges. 

If you feel that path is too divergent vs what you have proposed here let me know and I can start work on a seperate CHIP. It may be a good idea to do that regardless talks around impactful changes to farmers to elicit community sentiment at the same time.

reythia:
An accelerated timeline is perhaps justifiable with recent development, but terminating at a filter of 1 instead of 32 goes too far.

It's linearly punitive to farmers whilst having diminishing returns against plot grinding. Once the threshold of grinding being economically unviable is passed the downside risk is a reduction in honest nodes and netspace with no material upside vs plot grinding. 

Flash may well become the dominant storage within your referenced timelines (late 2020s) at the enterprise level based on new prices. I don't see it happening for home users and second hand markets anywhere close to that quickly, but that's my opinion.

You also mention 22TB drives being viable with filter 1 using uncompressed plots - but 22TBs are available today, and expected to reach much higher capacity over the next several years. I don't understand the security rationale for pushing high density HDDs out of honest netspace - doesn't that make grinding attempts more attractive not less?

Obsoleting HDDs prematurely also makes things like 'one click farming' and having millions of home users much more challenging aspirations. Whilst well intentioned, filter going to 1 seems it may hurt _overall_ security rather than help it.

SlowestTimelord:
> It's linearly punitive to farmers whilst having diminishing returns against plot grinding. Once the threshold of grinding being economically unviable is passed the downside risk is a reduction in honest nodes and netspace with no material upside vs plot grinding.

This is a great point! I hadn't thought about the diminishing gains at lower filter levels.

> Flash may well become the dominant storage within your referenced timelines (late 2020s) at the enterprise level based on new prices. I don't see it happening for home users and second hand markets anywhere close to that quickly, but that's my opinion.

I think we'll see it in waves as it may become an economical choice for high electricity rate regions.

> You also mention 22TB drives being viable with filter 1 using uncompressed plots - but 22TBs are available today, and expected to reach much higher capacity over the next several years. I don't understand the security rationale for pushing high density HDDs out of honest netspace - doesn't that make grinding attempts more attractive not less?

Pushing out high density HDDs is not the goal but highlighted as a detrimental impact of this approach that will need to be considered. Stopping at a lower filter level would alleviate it.

> Obsoleting HDDs prematurely also makes things like 'one click farming' and having millions of home users much more challenging aspirations. Whilst well intentioned, filter going to 1 seems it may hurt _overall_ security rather than help it.

I'd be curious what spare storage of home users in the future looks like -- even today overprovisioned storage is more likely to be on flash storage than hard drives.

Thank you for the thoughtful response, these are great call outs!

SlowestTimelord:
> The levelers that at the time were available to be pulled to provide ongoing security of the chain for CHIP 12 addressed what at the time seemed like a specific case around GH. We have a very different competing product to GH out now and indeed its impacts on NC have been dramatic for Chia. I think reassessing the levers pulled from a K minimum size need to be reevaluated. To be clear I am suggesting a change to the minimum K size is something that should be on the table. There are a wide variety of benefits that approach could have as well as a seperate set of challenges.

Changing minimum K-size is certainly on the table but I personally would like to consider it in the lens of preventing plot grinding. I don't see compressed farming as a problem unto itself but rather the specific pooling implementations of it which are separately addressed in CHIP-0022.

If a higher K-size is capable of exponentially increasing the costs of plot grinding then it could be worth the trade off of replotted netspace. I'm not convinced that it does though. Renting GPU clusters is as easy as ever (though maybe not currently cheap). I'd defer to someone that understands the requirements of higher k-size plotting to chime in. My fear is higher k-size makes Chia farming out of reach of common users.

> If you feel that path is too divergent vs what you have proposed here let me know and I can start work on a seperate CHIP. It may be a good idea to do that regardless. Involving farmers in talks around impactful changes to farmers and eliciting community sentiment at the same time sounds like a good approach.

Great point about eliciting more input from the farmers that will be impacted. A few vocal users in GitHub should not be making this decision for 90K+ farmers.

SlowestTimelord:
> An accelerated timeline is perhaps justifiable with recent development, but terminating at a filter of 1 instead of 32 goes too far.
> 
> It's linearly punitive to farmers whilst having diminishing returns against plot grinding. Once the threshold of grinding being economically unviable is passed the downside risk is a reduction in honest nodes and netspace with no material upside vs plot grinding.
> 
> Flash may well become the dominant storage within your referenced timelines (late 2020s) at the enterprise level based on new prices. I don't see it happening for home users and second hand markets anywhere close to that quickly, but that's my opinion.
> 
> You also mention 22TB drives being viable with filter 1 using uncompressed plots - but 22TBs are available today, and expected to reach much higher capacity over the next several years. I don't understand the security rationale for pushing high density HDDs out of honest netspace - doesn't that make grinding attempts more attractive not less?
> 
> Obsoleting HDDs prematurely also makes things like 'one click farming' and having millions of home users much more challenging aspirations. Whilst well intentioned, filter going to 1 seems it may hurt _overall_ security rather than help it.

I can't speak to the technical feasibility of the examples mentioned but some way of "converting" existing plots feels like a necessity for any proposal that forces a replot of 97%+ netspace such as k-size increase or plot format change.

danieljperry:
> During the research for CHIP-12 and CHIP-13, we spent considerable time discussing how aggressive the filter reduction needs to be. The economic model has remained constant, but now we have real-time data on GPU rental prices from runpod.io, vast.ai, etc. and some basic Bladebit plot times on an H100 (which is very important b/c it is a PCIe 5.0 card). I have updated the rental model to account for prices. The threat model was originally developed around commodity gaming cards that are more easily obtainable and have lower costs.
> 
> https://docs.google.com/spreadsheets/d/16K4Ki846npprN2q5OLGCbusCPX4Ci2Hlk_vMdrY2I4g/edit?usp=sharing
> 
> I have measured a 30 second plot with 15 second phase one, a single H100. If the table 7 finalization is removed, which is not needed for grinding, that shaves off another few seconds. Today plot grinding could be run on a single card, but it would be incredibly stupid as those instances are renting for around $4/hr.
> 
> To prevent plot grinding, the profitability needs to be some multiple of honest farming profitability to ensure it doesn't become viable during short-term price volatility. When the XCH price goes up, we want to ensure adequate time for honest farmers to get the space online. CHIP-13 does this and gives a tremendous amount of room. 5x is more than enough, meaning if the price increases by 5x and plot grinding still aren't profitable, it will never happen because the honest space will beat it out and increase the Netspace,
> 
> I would, however, like to see all VRAM POC of Bladebit, as cards with a tremendous amount of high-speed memory are being developed for AI training. H200 will have 141GB and 4.8TB/s memory bandwidth, but actually less FP32 performance than a 4090. This is meaningful because the Chia proof of space is a "memory bandwidth bound" workload.
> 
> Surprisingly, plot compression has helped defend against plot grinding due to the higher effective Netspace.
> 
> To reiterate what was mentioned above, although I believe SSDs transition will happen even sooner than most industry analysts, I don't think the Chia protocol should restrict hard drives from being viable for farming.
> 
> To summarize, I recommend getting CHIP-13 or replacement back online immediately. This will be more than enough to deter plot grinding during short-term XCH price volatility, and god forbid, the next bull market.

Thanks for this excellent analysis!

SlowestTimelord:
Thank you for chiming in with expertise and data, @jmhands!

> To reiterate what was mentioned above, although I believe SSDs transition will happen even sooner than most industry analysts, I don't think the Chia protocol should restrict hard drives from being viable for farming.

At what filter size is HDD farming not viable? I thought plot filter removal would just barely make the large capacity slow drives not viable. (e.g. 20TB 5400rpm was the example that was at the cusp IIRC).

SlowestTimelord:
Thank you all for the feedback and discussions (both here and other channels). I am withdrawing this CHIP and will proposed a modified version after the first plot filter reduction to observe how netspace and farmer experience responds.

PR #96, opened 2024-01-16
fix broken links, update others to static

BrandtH22:
original links used dynamic references to the main branch, this updates those to static links for the respective docs eliminating the potential of broken links in the future.

This also fixes other broken links


hoffmang9 (dismissed):

danieljperry (changes requested):
Thanks for updating these links. Could you add an Errata section with a brief explanation of what you did? You can follow this CHIP's errata as a reference: https://github.com/Chia-Network/chips/blob/main/CHIPs/chip-0005.md#errata

danieljperry (approved):

PR #116, opened 2024-04-26
CHIP-0034: keccak and base64 operators

danieljperry:


freddiecoleman:
Thanks, @danieljperry ! This CHIP has been assigned `CHIP-34` and is now a `Draft`. Feel free to leave a review here, or to discuss it in the #chips channel of our [Discord](https://discord.gg/chia).

greimela:
Great addition!

Regarding passkey support: the passkey standard uses the urlencoded version of base64, without padding. So those two aspects could be configurable via arguments, or via separate operators.

CHIPs/chip-0034.md
```diff
@@ -0,0 +1,129 @@
...
+CLVM Cost: `[todo]` base, `[todo]` per argument, `[todo]` per byte
+
+### `base64_encode`
+
+Opcode: [todo]
+
+Functionality: Calculate and return the Base64 encoding of the input atom(s)
+
+Arguments:
+* If zero arguments, the result will be a successful no-op
```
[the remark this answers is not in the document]
danieljperry: I think it should be the Base64 encoding of an empty string (`NIL`)

CHIPs/chip-0034.md
```diff
@@ -0,0 +1,129 @@
...
+* If one argument, the Base64 encoding of the argument will be returned
+* If two or more arguments, each argument will be concatenated, and the Base64 encoding of the result will be returned
```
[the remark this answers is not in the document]
danieljperry: Correct, I'll update it.

CHIPs/chip-0034.md
```diff
@@ -0,0 +1,129 @@
...
+### `base64_decode`
+
+Opcode: [todo]
+
+Functionality: Calculate and return the Base64 decoding of the input atom(s)
+
+Arguments:
+* If zero arguments, the result will be a successful no-op
+* If one argument, the Base64 decoding of the argument will be returned
+* If two or more arguments, each argument will be concatenated, and the Base64 decoding of the result will be returned
```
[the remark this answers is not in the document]
danieljperry: Single argument is fine.

CHIPs/chip-0034.md
```diff
@@ -0,0 +1,129 @@
...
+CLVM Cost: `[todo]` base, `[todo]` per argument, `[todo]` per byte
+
+### `base64_decode`
+
+Opcode: [todo]
+
+Functionality: Calculate and return the Base64 decoding of the input atom(s)
+
+Arguments:
+* If zero arguments, the result will be a successful no-op
```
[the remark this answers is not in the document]
danieljperry: yes

CHIPs/chip-0034.md
```diff
@@ -0,0 +1,129 @@
...
+
+Usage: `(base64_encode A B …)`
+
+CLVM Cost: `[todo]` base, `[todo]` per argument, `[todo]` per byte
+
+### `base64_decode`
+
+Opcode: [todo]
+
+Functionality: Calculate and return the Base64 decoding of the input atom(s)
```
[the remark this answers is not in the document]
danieljperry: Thanks, I'll update it.

CHIPs/chip-0034.md
```diff
@@ -0,0 +1,129 @@
...
+* The operators to be added are unlikely to be contentious. However, as with all forks, there will be a risk of a chain split.
+* The soft fork could also fail to be adopted. This might happen if an insufficient number of nodes have upgraded to include the changes introduced by this CHIP prior to the fork's block height.
+
+The operators will be introduced in multiple phases:
+* **Pre-CHIP**: Prior to block `[todo]`, any attempt to call the new operators will result in a successful no-op.
+* **Soft fork**: A soft fork will activate at block `[todo]`. From that block forward, the new operators will exhibit the functionality laid out in this CHIP. They will need to be called from inside the `softfork` guard. 
+* **Hard fork** (hypothetical): The Chia blockchain currently does not have any planned hard forks. However, in the event that a hard fork is ever enacted, the operators from this CHIP could be added to the core CLVM operator set. If this were to happen, the operators could be also be called from outside the `softfork` guard. (Note that the operators would still be callable from inside the `softfork` guard if desired.)
```
Rigidity, line 55: We have the plot filter hard fork activation upcoming & the plot format v2 hard fork planned; does this mean "no CHIPs for hard forks which are not yet released"?

Rigidity (commented):
Looks good so far

danieljperry:
Thanks for the comments! I have updated the CHIP accordingly.

CHIPs/chip-0034.md
```diff
@@ -36,23 +36,23 @@ CLVM currently includes an atomic operator called [`sha256`](https://chialisp.co
 
 This CHIP will add an atomic operator to the CLVM called `keccak256`, which will calculate and return the keccak-256 hash of the input atom(s). The primary reason to add this operator is to support Ethereum addresses, which also rely on keccak-256.
 
-This CHIP will also add operators to the CLVM called `base64_encode` and `base64_decode`, which will calculate and return the Base64 encoding/decoding of the input atoms(s). The primary reason to add the `base64_encode` operator is to support passkeys that rely Base64 encoding. `base64_decode` is being added for completeness.
+This CHIP will also add operators to the CLVM called `base64_encode` and `base64_decode`, which will calculate and return the Base64 encoding/decoding of the input atoms(s). The primary reason to add the `base64_encode` operator is to support passkeys that rely on Base64 encoding. `base64_decode` is being added for completeness.
```
arvidn, line 39: Assuming I found the right specification for passkeys, could we have references:
https://passkeys.dev/docs/reference/specs/
https://www.w3.org/TR/webauthn-2/

CHIPs/chip-0034.md
```diff
@@ -62,6 +62,10 @@ While the Keccak-256 and Base64 operators are not related, they each require sof
 
 Each of the new operators will incur a CLVM cost, as detailed below. If this CHIP is adopted, the new operators will be optional when designing Chia coins.
 
+The Base64 alphabet used in this CHIP is specified in [Section 4 of the RFC 4648 Standard](https://www.rfc-editor.org/rfc/rfc4648.html#section-4).
```
arvidn, line 65: When following the passkey specification to webauthn-2 ( https://www.w3.org/TR/webauthn-2/ ), it clearly specifies the encoding as `base64url`, which is not section-4 that you reference above. It's section-5.

"Base 64 Encoding with URL and Filename Safe Alphabet"

The RFC calls this encoding `base64url`. Perhaps we should consider naming the functions this as well.

CHIPs/chip-0034.md
```diff
@@ -62,6 +62,10 @@ While the Keccak-256 and Base64 operators are not related, they each require sof
...
+
+Each Base64 character is encoded with six bits, and decoded to eight bits. Therefore, the ending of some Base64 output strings [require padding](https://en.wikipedia.org/wiki/Base64#Output_padding). The padding used in this CHIP is `=`, which is also specified in the RFC 4648 Standard.
```
arvidn, line 67: the webauth-2 also specifies that padding characters are *omitted*.

From the specification:

> The term Base64url Encoding refers to the base64 encoding using the URL- and filename-safe character set defined in Section 5 of [[RFC4648]](https://www.w3.org/TR/webauthn-2/#biblio-rfc4648), with all trailing '=' characters omitted (as permitted by Section 3.2) and without the inclusion of any line breaks, whitespace, or other additional characters.

CHIPs/chip-0034.md
```diff
@@ -71,9 +75,10 @@ Opcode: [todo]
 Functionality: Calculate and return the Keccak-256 hash of the input atom(s)
```
arvidn, line 75: How does keccak256 differ from SHA3-256?
I take it since we say "keccak-256", we don't mean SHA3-256, but the version before winning the NIST competition. Based on this: https://crypto.stackexchange.com/questions/15727/what-are-the-key-differences-between-the-draft-sha-3-standard-and-the-keccak-sub

CHIPs/chip-0034.md
```diff
@@ -86,9 +91,10 @@ Opcode: [todo]
 Functionality: Calculate and return the Base64 encoding of the input atom(s)
 
 Arguments:
-* If zero arguments, the result will be a successful no-op
-* If one argument, the Base64 encoding of the argument will be returned
-* If two or more arguments, each argument will be concatenated, and the Base64 encoding of the result will be returned
+* If zero arguments, the result will be the Base64 encoding of an empty string (`NIL`), in other words, the result will be `NIL`
+* If one or more arguments:
+  1. Each argument (if more than one) will be concatenated
+  2. The Base64 encoding (following [Section 4 of the RFC 4648 Standard](https://www.rfc-editor.org/rfc/rfc4648.html#section-4), and padded with `=` when necessary) of the resulting string will be returned
```
arvidn, line 97: this should refer to section-5 and it should state that no padding is produced

CHIPs/chip-0034.md
```diff
@@ -98,14 +104,16 @@ CLVM Cost: `[todo]` base, `[todo]` per argument, `[todo]` per byte
 
 Opcode: [todo]
 
-Functionality: Calculate and return the Base64 decoding of the input atom(s)
+Functionality: Calculate and return the Base64 decoding of the input atom
```
arvidn, line 107: I think it's more straight-forward to just say "return the decoded form of the base64 input string".

CHIPs/chip-0034.md
```diff
@@ -98,14 +104,16 @@ CLVM Cost: `[todo]` base, `[todo]` per argument, `[todo]` per byte
...
 
 Arguments:
-* If zero arguments, the result will be a successful no-op
+* If zero arguments, the result will be an empty string (`NIL`)
 * If one argument, the Base64 decoding of the argument will be returned
-* If two or more arguments, each argument will be concatenated, and the Base64 decoding of the result will be returned
+* If more than one argument, an exception will be raised
+
+Note that the argument must be in Base64 format, encoded according to [Section 4 of the RFC 4648 Standard](https://www.rfc-editor.org/rfc/rfc4648.html#section-4). It must be padded with `=`, also according to the RFC 4648 Standard. If the argument is not encoded in the proper format, or if it is not properly padded, an exception will be raised.
```
arvidn, line 114: Same thing here. section-5, padding is disallowed, any characters not in the encoding alphabet are disallowed. e.g. new lines

CHIPs/chip-0034.md
```diff
@@ -98,14 +104,16 @@ CLVM Cost: `[todo]` base, `[todo]` per argument, `[todo]` per byte
 
 Opcode: [todo]
 
-Functionality: Calculate and return the Base64 decoding of the input atom(s)
+Functionality: Calculate and return the Base64 decoding of the input atom
 
 Arguments:
-* If zero arguments, the result will be a successful no-op
+* If zero arguments, the result will be an empty string (`NIL`)
```
arvidn, line 110: it's not clear to me that we would want to support 0 arguments. Typically we only do that for operators that take a variable number of parameters

danieljperry:
Thanks for the reviews, I believe I have addressed all of the concerns raised.

CHIPs/chip-0034.md
```diff
@@ -0,0 +1,141 @@
...
+* `base64url_decode` -- for Base64 decoding; assumes the input uses Base64 encoding with a URL- and filename-safe alphabet
+
+The operators will be accessible from behind the `softfork` guard. Hypothetically, in the event of a planned hard fork, the operators could be added to the core CLVM.
+
+## Definitions
+
+Throughout this document, we'll use the following terms:
+* **Chialisp** - The high-level [programming language](https://chialisp.com/) from which Chia coins are constructed
+* **CLVM** - [Chialisp Virtual Machine](https://chialisp.com/clvm), where the bytecode from compiled Chialisp is executed. Also commonly refers to the compiled bytecode itself
+* **Keccak-256** - The same encryption standard Ethereum uses. A few notes on this standard:
```
Yakuhito, line 30: The same **hashing** standard Ethereum uses.

danieljperry:
We will hold a public zoom call to talk about this CHIP on May 20 at the following time:
7 AM PDT / 2 PM UTC / 10 PM China
See the #chips channel of our public [Discord](https://discord.com/invite/chia) for more info.

CHIPs/chip-0034.md
prozacchiwawa: https://github.com/prozacchiwawa/elm-keccak/blob/bccbc3ed87664b9cae36c18d6520cf742478dec8/src/Keccak.elm#L724 <-- specifically how they differ, just in the padding byte value.

CHIPs/chip-0034.md
```diff
@@ -0,0 +1,183 @@
...
+
+Arguments:
+
+Exactly one argument is required. If there is not exactly one argument, an exception will be raised.
+
+The argument must be in `base64url` format, encoded according to [Section 5 of the RFC 4648 Standard](https://www.rfc-editor.org/rfc/rfc4648.html#section-5). It must not be padded with `=` or any other characters. If the argument is not encoded in `base64url` format, or if it includes any padding, or if it includes any additional characters that are not part of the `base64url` alphabet, an exception will be raised.
+
+Usage: `(base64url_decode A)`
+
+CLVM Cost: `[todo]` base, `[todo]` per argument, `[todo]` per byte
```
richardkiss, line 122: What's the motivation for base64 encoding & decoding? I understand it's something to do with ETH. Where can I find more details?

Including base64 data instead of the corresponding binary data on chain seems like a mistake, since reading base64 data would come from a computer before being dumped on chain, and there are plenty of opportunities for the reading client to decode before dumping on chain. Maybe ETH does this, due to some bad design somewhere. But I worry we're proposing changes to clvm to make simpler interoperability with bad design on ETH or elsewhere, which just encourages people to repeat those design errors on chia.

hoffmang9: base64 is in support of passkey support.

Rigidity: If a signature must be validated against a base64 message, but you need to validate the message matches something within the puzzle, you have two options:
* Give the puzzle the encoded base64 message and decode it, then check it matches
* Give the puzzle the original message, check it matches, then base64 encode it before validating the signature

Both of these require base64 opcodes

Rigidity: Keccak is for Ethereum address compatibility - it allows an application to construct a Chia puzzle that is only spendable by whoever owns the private key corresponding to the Ethereum address. Which means you can send XCH to an Ethereum address with a little bit of wrapping behind the scenes

Rigidity: And base64 allows you to have a puzzle that can only be spent by including a signature created with a passkey (meaning you don't need to have access to the original secret key, just the signer)

richardkiss: > If a signature must be validated against a base64 message, but you need to validate the message matches something within the puzzle, you have two options:
> 
> * Give the puzzle the encoded base64 message and decode it, then check it matches
> * Give the puzzle the original message, check it matches, then base64 encode it before validating the signature
> 
> Both of these require base64 opcodes

Why include and sign a base64 message when you could instead include and sign the (shorter) binary data it represents?

(A possible answer is "there is no good reason for new contracts, but we want to interoperate with existing ETH contracts that do this.")

greimela: It doesn’t have anything to do with ETH.

When passkeys are used via the Web Authentication API, they always encode the challenge (message) with base64 before signing it.

That’s why we need to encode the message in the puzzle too, so we can validate the passkey signature.

richardkiss: @greimela Appreciate the response. Do you have a reference where I can learn more about this? It seems surprising to me that they couldn't sign binary data directly and I'm wondering what possible motivation there could be.

arvidn: I posted some links here: https://github.com/Chia-Network/chips/pull/116#discussion_r1586337354 and here: https://github.com/Chia-Network/chips/pull/116#discussion_r1586330975

greimela: Yes, thanks, those are the ones!

Yakuhito:
Hi! Wanted to put [this](https://x.com/yakuh1t0/status/1792472344427905034) here. This CHIP could potentially allow Chia to integrate with any wallets that support ETH - including hardware wallets - by 'impersonating' a custom EVM chain during signing. From my (very brief) research, I think `RLP encoding` should be looked into - again, this is just a raw idea.

danieljperry:
CHIP-34 is now in `Review`. Please review as time permits. We seem to have consensus on adding the three operators from this CHIP, if there are no objections from the community, we will move it to `Last Call` soon.

danieljperry:
In case you missed it, here is our discussion of this CHIP:
https://youtu.be/hFwPBfM4s8c

CHIPs/chip-0034.md
richardkiss: I've looked at this more today and I think I understand this better now. WebAuthn apis sign JSON messages of a certain format, ie. certain keys must be present and must have specific formats, so that you don't sign arbitrary messages with arbitrary keys. For details, see https://www.w3.org/TR/webauthn-2/#client-data

So the thing that's signed is a JSON blob which includes a `challenge:` key, which is server-determined. In the prototype, we hack it to be a hash of the target conditions for the coin. Validating requires checking the signature, which is against the entire JSON blob, and then extracting the portion of the JSON blob (via an offset hint) that points to the base64 data, and comparing it against the delegated puzzle.

danieljperry:
We have decided to withdraw CHIP-34 from consideration. The reasons for this are threefold:

1. There isn't sufficient demand for the keccak256 operator to justify forking the consensus in order to implement it.

2. For base64 encoding, we are able to do this without adding a new operator. While it is true that prior to the activation of the 2.1 hard fork, the operator from this CHIP would have been significantly less costly than the non-CHIP method, this is a moot point because this CHIP would not have activated prior to the 2.1 fork's activation. 

After the 2.1 fork takes full effect, block compression will be enabled. At this point, the operator from this CHIP would only have been about 2.3% more efficient in CLVM cost versus the non-CHIP version of base64 encoding. We deemed that this saving would not justify forking the consensus.

3. For base64 decoding, this was only ever added for completeness. There is not currently a use case for it.

If we ever do decide to implement these operators in the future, the source material will still be available for a new CHIP.

PR #138, opened 2024-12-17
CHIP-0039: Fee Service Standard

freddiecoleman:


danieljperry:
This CHIP is a `Draft`. It proposes a standard way for a service to include fees with transactions. Feel free to leave a review here, or to comment in the #chips channel of [our Discord](https://discord.gg/chia).

Yakuhito:
Great CHIP! I can see some situations where services might need a few mojos to start a transaction. For example:
 - Minting singletons (e.g., vaults/NFTs) requires 1 mojo
 - Minting CATs currently requires 1000 mojos/CAT
 - A 'subset' of minting CATs is finishing bridging operations, where one needs to provide the mojos needed to mint the wrapped tokens

Would it make sense to have the same services be able to provide a one-sided offer to get the mojos required to build transactions with? This is what I did for the [warp.green welcome kits](https://github.com/Yakuhito/welcome-kits/). The fee could very well still be attached at the end (when the final spend bundle is available), but could also be included in the initial offer if the app knows what fee it'll need. That way, you also remove some trust in the service, as the dApp itself can push the spend bundle to the mempool.

danieljperry:
> Great CHIP! I can see some situations where services might need a few mojos to start a transaction. For example:
> 
>     * Minting singletons (e.g., vaults/NFTs) requires 1 mojo
> 
>     * Minting CATs currently requires 1000 mojos/CAT
> 
>     * A 'subset' of minting CATs is finishing bridging operations, where one needs to provide the mojos needed to mint the wrapped tokens
> 
> 
> Would it make sense to have the same services be able to provide a one-sided offer to get the mojos required to build transactions with? This is what I did for the [warp.green welcome kits](https://github.com/Yakuhito/welcome-kits/). The fee could very well still be attached at the end (when the final spend bundle is available), but could also be included in the initial offer if the app knows what fee it'll need. That way, you also remove some trust in the service, as the dApp itself can push the spend bundle to the mempool.

This is certainly something that could be added to a fee paying service, but I'm not sure if it belongs in the CHIP. Let's plan to have a public zoom call to discuss this, and other ideas pertaining to the CHIP, early in 2025.

danieljperry:
We had a discussion of this CHIP, with several ideas for improving it and implementing it.
https://youtu.be/kWXT4WXtkC4

PR #143, opened 2025-02-04
Protected Single Sided Offers

Rigidity:


danieljperry:
This CHIP is now a `Draft`. Please leave your reviews here.

danieljperry:
My biggest concern currently is that the offer maker has to rely on the taker to use a wallet that implements this CHIP. Is there a way to shift the burden of protecting the offer so that it doesn't fall entirely on the maker?

We should put together a Zoom call to discuss this and other questions.

Rigidity:
> As for the invoice solution, it sounds more like wallets using the Offer File format as a means of auto-filling details of a transaction. If that's the case, has there been thought given to a more generic schema/format that invoice creators could instead leverage? Like a manual copy/paste version of WalletConnect functionality.

I realize now that invoice offers as they work today can actually be secure (assuming the wallet asserts its payments correctly). So I've changed the direct payment method to a suggestion rather than requirement, and also mentioned other options are possible (though out of the scope of this CHIP).

I agree that a deep link or smaller string for invoices would make more sense from a usability perspective, though I should note that invoice offers at a Chialisp level support requesting specific memos be sent to the address as well, so you could differentiate payments with either solution.

Rigidity:
> Securely adding fees at the same time you accept the offer seems like it would accomplish the same thing. The thought there is probably that single-sided offer acceptors don't have XCH for fees, right?

Yeah that's a good point but one main use case here is for people who just installed the wallet and scan a QR code or NFC and don't have any assets already.

But I can update the CHIP to mention fee spends as an alternative, and implement that in Sage if the user adds a fee. Since it would save on cost.

danieljperry:

danieljperry:
The recording from our discussion of this CHIP is here:
https://youtu.be/lqHeQOP_Fuw

judeallred (approved):

PR #149, opened 2025-03-05
Condition Options Puzzle

Rigidity:


Yakuhito:
I would add that this is a simplified [p2_1_of_n](https://github.com/Chia-Network/chia_puzzles/blob/main/puzzles/p2_1_of_n.clsp) that uses conditions instead of a quoted delegated puzzle + NIL delegated solution. I'd be curious which one is more efficient assuming `simplify_merkle_proof` and `sha256tree` are compressed.

Rigidity:
Agreed, I've switched the implementation to `p2_1_of_n`.

PR #150, opened 2025-03-05
CHIP-0044: Clawback Standard v2

Rigidity:


Yakuhito:
Awesome! I like the idea that the receiver wallet can treat the coins as 'owned' after the delay, even without spending the clawback coin (which only needs to be spent when the value being sent is needed for a transaction - in the same transaction)

In the code,  I see that the memo format is `sender_puzzle_hash seconds`. I would like to propose `receiver_puzzle_hash sender_puzzle_hash seconds` instead, as that:
 * would allow wallets to be notified when they're the recipient of a clawback transaction, since the hint becomes `receiver_puzzle_hash`
 * would allow blockchain observers to fully build the clawback puzzle from the parent spend (without the receiver having to send their puzzle hash)
 * would support CATs and NFTs, since `receiver_puzzle_hash` would be the hash of the 'custody' (inner) puzzle, which gets wrapped by the outer puzzles (and you can still add a memo to the `CREATE_COIN`s only if the clawback is for CATs/NFTs, which anyone can determine knowing the three values)

Rigidity:
> Awesome! I like the idea that the receiver wallet can treat the coins as 'owned' after the delay, even without spending the clawback coin (which only needs to be spent when the value being sent is needed for a transaction - in the same transaction)

Well, this is true but it would require spending the coin twice do to current limitations (the receiver isn't part of the inner puzzle and so would rather have to push it through to themselves and then spend it).


> In the code,  I see that the memo format is `sender_puzzle_hash seconds`. I would like to propose `receiver_puzzle_hash sender_puzzle_hash seconds` instead

The hint goes before the proposed memo structure, although you raise a good point which is that the hint should always be included so the receiver can find the coin, even for XCH.

Rigidity:
I've clarified the intended memo structure in the CHIP

Rigidity:
I've updated the CHIP to include the revised implementation of clawback v2, which uses a `p2_1_of_n` puzzle rather than `p2_condition_options`. This means that only the clawback coin needs to be spent, rather than an additional coin which sends a message to prove ownership. Additionally, it only reveals the spend path that's needed on-chain, saving on CLVM cost. All of the functionality is the same, and more tests have been written.

danieljperry:
This CHIP is now a `Draft` and has been assigned CHIP-44. It describes how to implement a new form of clawback coins by using `ASSERT_BEFORE`. Wallet devs will likely be interested in this standard. Please leave your reviews here, and feel free to discuss in the #chips channel of [our Discord](https://discord.gg/chia).

richardkiss (commented):
There's a risk here in that the old standard, you could send money to pay a debt and not have to worry about a typo in the address. If you brick your coin to an unspendable public key, you can come back in a day or a year or a century and claw it back, as long as it hasn't been "clawed forward" by anyone.

This new standard has a finite time for clawback, even in the case of a bricked coin. You might consider coming up with a name that can't be confused with the unbrickable "clawback". Maybe "clawback_until" or "vesting_payment" or something.

danieljperry:
We will discuss this CHIP in a public Zoom call on April 3 at 10 AM EDT. See the #chips channel in our Discord for more info.

danieljperry:
Here is the recording of the discussion for this CHIP:
https://youtu.be/-kGkp7lXWJQ

danieljperry:

danieljperry:
This CHIP is now in `Last Call`. If no further changes are required in the next two weeks, then it will be moved to `Final` status.

danieljperry:

PR #160, opened 2025-05-19
CHIP-0048: New Proof of Space

danieljperry:
We propose a new Proof of Space protocol for the Chia blockchain that addresses weaknesses in energy efficiency, rental attack resistance, and plot format stability.


BrandtH22:
Note that timelines associated with integrating PoS 2.0 can be found in [CHIP 49](https://github.com/Chia-Network/chips/pull/161)

danieljperry:
The term `plot difficulty` could be confused with the network's global `block difficulty`. In addition, while "difficulty" in this context describes the amount of time required to create a _plot_, it could easily be confused for the time required to create/solve a _proof_. With these considerations in mind, we have decided to update the term `plot difficulty` to `plot strength`.

danieljperry:
This CHIP is now in `Review` status. Please leave your reviews here, and feel free to discuss it in the #chips channel of [our Discord](https://discord.gg/chia).

trgarrett:
Could you please include in the supporting documents for this CHIP a cost model for attacking PoS2 with AES ASICs? I understand the barrier imposed to GPU-based rental attacks, but that seems to incompletely capture the attack surface.

Edit: To clarify my question, I don't think there is any threat there, but I think it's important to recognize the hypotheticals to rule them out.

danieljperry:
> Could you please include in the supporting documents for this CHIP a cost model for attacking PoS2 with AES ASICs? I understand the barrier imposed to GPU-based rental attacks, but that seems to incompletely capture the attack surface.

I'll pass this request on to Dr. Nick.

drnick23:
> Could you please include in the supporting documents for this CHIP a cost model for attacking PoS2 with AES ASICs? I understand the barrier imposed to GPU-based rental attacks, but that seems to incompletely capture the attack surface.

CPU's with hardware support for AESENC are, as far as I understand, already AES ASIC's (hence the 10x speed-up over software implementation).  If GPU's also had in-built AESENC hardware support this could speed up hash rate, and likely turn the gpu compute-bound hash into memory bound in the worst-case.

I could add an attack estimation based on the AES hashing portion becoming memory bound on the GPU as an ASIC attack. 

Trying to guess what a potential ASIC fully built for plotting/attacking could be estimated by inputting expected hash rate and memory bandwidth for the ASIC, but I would not know what those performance differences vs a 5090 could be or the cost to build it per unit.

trgarrett:
> CPU's with hardware support for AESENC are, as far as I understand, already AES ASIC's (hence the 10x speed-up over software implementation).

A general purpose computing device is by definition not an ASIC. The Chia team commissioned a VDF ASIC. It seems like you should have the knowledge base in-house to at least get an estimate of performance and cost. From the research I've done, it looks like AES ASICs may be 50 to 100 times faster than even extension-enhanced CPUs.

Edit: the optimization progression in all PoW chains was: CPU, GPU, FPGA, ASIC. The pattern has been repeated many times.

hoffmang9:
@trgarrett I'm borrowing this from Sonnet 4.6 but I've reviewed it and it is correct:

# AES ASIC vs On-Chip AES: Feasibility of Material Speed Improvement

This is a fun question that gets at some deep hardware realities. The short answer is: **extremely difficult to improve meaningfully, and essentially impossible at the same voltage/process node.**

## Why On-Chip AES Is Already Nearly Optimal

Modern x86 (AES-NI), ARM (AES extensions in ARMv8), and Apple Silicon all implement AES-128/256 in dedicated silicon. The key characteristics:

- **Single-cycle or near-single-cycle latency** for a full AES round, with pipelining giving ~1 block/cycle throughput in optimized modes (e.g., CTR, GCM)
- **AES-GCM throughput on modern Intel** (e.g., Sapphire Rapids) can hit ~100 Gbps per core using VAES + VPCLMULQDQ with 8-way parallelism across 512-bit vectors
- The critical path through SubBytes + ShiftRows + MixColumns + AddRoundKey is already reduced to a handful of gate delays using composite field arithmetic for the S-box

## What an ASIC Could Theoretically Do

1. **Massive parallelism** — an ASIC could instantiate thousands of AES engines. But this is throughput scaling, not latency improvement, and a CPU core can already do this with VAES on wide vectors. A GPU does this too.

2. **Remove everything else** — no branch predictors, caches, out-of-order logic. But those don't sit in the AES datapath anyway; AES-NI is already a bolt-on functional unit.

3. **Custom cell libraries / aggressive pipelining** — you could cut latency slightly with deeper pipelining and faster standard cells, but you're fighting the same physics.

## The Fundamental Bottleneck: The AES S-Box

The irreducible core is the GF(2⁸) multiplicative inverse for SubBytes. In silicon, the best implementations use a tower field decomposition (GF(2⁴) over GF(2²) over GF(2)) which reduces the critical path to roughly:

- ~4-5 XOR/AND gate delays for the inversion
- Plus linear transformations

At 3-5nm nodes, that's already sub-200ps. You're not going to get orders of magnitude out of this — maybe 20-30% with exotic cell libraries and aggressive Vt tuning, which is well within what CPU vendors already do on their AES units.

## Bottom Line

For **AES specifically**, the CPU vendors have essentially closed the gap. An AES ASIC at the same process node would offer:
- Similar or marginally better latency (single-digit percent improvement)
- Higher aggregate throughput only through sheer parallelism — but at that point you're comparing it to a many-core CPU or GPU, which can also parallelize

The only realistic niche is in constrained IoT/embedded contexts where you want AES with minimal power and area, not speed — and even there, ARM's TrustZone Cortex-M cores with hardware AES are already extremely efficient.

The economics are brutal: by the time you tape out an AES ASIC, Intel or Apple has shipped another generation that makes the delta irrelevant.

trgarrett:
@hoffmang9 I've spent quite a few tokens in Opus 4.6 in thinking mode to probe this topic. It looked at it from a number of different angles, not all of which I find useful. For example, I recognize that farming a challenge faster with an ASIC poses very little real benefit. However, I wanted to provide its full output for the CHIP in case there are any viable paths that merit further discussion
[ASIC_ATTACK_ANALYSIS.pdf](https://github.com/user-attachments/files/26143575/ASIC_ATTACK_ANALYSIS.pdf).

hoffmang9:
How much are the one time costs to build such an ASIC? Does it assume 2nm or 3nm which will take 24 months at best? I bet the one time cost approaches $50M but that then runs right back into the next generation of CPU optimized AES...

trgarrett:
> How much are the one time costs to build such an ASIC? 

One scenario I saw proposed was $10-50M in R&D and manufacturing for a $3k-5k unit that simulated 9 plots. That would definitely be a poor bet. I didn't think ASICs would beat storage today in any comparison. I was mainly curious how many orders of magnitude apart they might be and how long they might be safely ignored. Even if you had a magic button to fab ASICs for free, they don't seem remotely close to viable. That's cool to see.

danieljperry:
Nick has made some updates to this CHIP with the latest commit:
https://github.com/Chia-Network/chips/pull/160/commits/783feb75f46c1ed5828410c21120f54bd0a45488

Please review this update (and the rest of the CHIP if you haven't done so already).

vertor222:
https://claude.ai/public/artifacts/89ace036-fa68-4c4c-93aa-72088364a7d9

danieljperry:
> https://claude.ai/public/artifacts/89ace036-fa68-4c4c-93aa-72088364a7d9

What is this?

trgarrett:
I have not found a way to attack this, but it is a difference between behavior at different k levels. I have run several attacks at this that seemingly reduce down to noise, but since k values below 22 and above 22 behave differently, I wanted to raise it for discussion.

Claude artifact:

At k=28, the FeistelCipher's get_round_key requests 84 bits (3×28) of key material per round but stores the result in a uint64_t, silently discarding the top 20 bits during the byte-accumulation loop in slice_key. When feistel_round then splits the 64-bit round key into three 28-bit sub-keys (b, c, d), the third sub-key d — extracted via round_key >> 56 — has only 8 real bits of key material; its upper 20 bits are always zero. This affects all 4 rounds identically: each round's d sub-key carries 8 bits of entropy instead of 28. Because encrypt and decrypt use the same truncated keys, round-trip consistency is preserved — which is why existing tests pass and the bug has remained latent.

The practical consequence is that the Feistel cipher operates as a weaker permutation than designed. The d sub-key feeds directly into the ChaCha-inspired quarter-round (d ^= a; d = rotate_left(d, 16, k)) whose output propagates through c and b and ultimately determines the final ciphertext half. With 20 of its 28 bits pinned to zero across every round, the cipher mixes fewer bits of the plot_id into the output — the effective key diversity per round is reduced by roughly 60 bits (20 bits × 3 sub-key-dependent operations). This matters because the Feistel cipher is the mapping that produces proof fragments (quality strings), which are hashed against the difficulty target to determine block winners. A permutation with reduced diffusion is more likely to exhibit statistical non-uniformities — subtle biases in the fragment distribution, higher collision rates, or structural patterns — that could give certain plot_ids a measurable statistical advantage. At k=28, the most common production plot size, every plot on the network is using this degraded permutation.

danieljperry:
> https://claude.ai/public/artifacts/89ace036-fa68-4c4c-93aa-72088364a7d9

FYI, Claude is not available globally. I cannot view this easily, which is why I asked what it is. In the future, instead of dropping a link that many people won't be able to view (and others won't click because they are unsure if it is malicious), please include at least a summary of the link's contents, or include a complete write-up in your comment. For bonus points, include a link that is viewable globally.

vertor222:
> > https://claude.ai/public/artifacts/89ace036-fa68-4c4c-93aa-72088364a7d9
> 
> FYI, Claude is not available globally. I cannot view this easily, which is why I asked what it is. In the future, instead of dropping a link that many people won't be able to view (and others won't click because they are unsure if it is malicious), please include at least a summary of the link's contents, or include a complete write-up in your comment. For bonus points, include a link that is viewable globally.

All complaints to anthropic! I don't trust them too ;)...

PR #161, opened 2025-05-19
CHIP-0049: 3.0 Fork Info

danieljperry:
Provide all necessary timelines for the hard fork to support the new proof format.


Yakuhito:
From my point of view, there are two 'phases' of the timeline discussed in this CHIP:
 * **Technical Integration & Testing**, which finishes once Chia 3.0 is released in Q4 2025.
 * **Hard Fork Activation & Phased Activation of the New Format**: The hard fork is currently set to activate at/after block height 8.800.000 (around 2026-06-01). The phase-out period would then take an additional 256 days, where the 'weight' of old plots lowers while that of plots using the new format increases.

I do not have an opinion regarding the timing of the first stage. CNI has an amazing blockchain team, and I'm confident the proposed changes will go through extensive testing before Chia 3.0 is rolled out. Security of the transition and the new format must come on the first place.

That being said, I have several arguments for significantly shorter hard fork activation heights & phase-out period:
 * **Farmers have a clear reason to upgrade their nodes**: The `keccak256` softfork took around six months to be activated after a version that supported it was released. That makes sense, as node operators are only loosely incentivized to update their software to a newer version that supports the soft fork. With a new Proof of Space, however, all farmers will need to update to Chia 3.0 to be able to start creating their new plots. Additionally, the problem of upgrading a Chia node is one of initiative - it likely only takes a few minutes to complete the task. Chia 3.0 has a very strong 'why upgrade your node right now' narrative for farmers.
 * **Even with a lower netspace, the network is still secure**: This argument has been frequently brought up when discussing declining netspace. I think the phase-out should be long enough to let most farmers to replot their farms - but we should treat farmers with very big capacities and few graphics cards as an outlier. In [this video](https://youtu.be/cFf_mjuTqcY), Dr. Nick mentioned that a 3060 can probably generate up to 40TB worth of plots per day. That translates to 1.2 PB (1200 TB) of new plots generated over the span of 30 days. Moreover, bigger farms already likely have more than one graphics card to allow them to farm compressed plots at the moment, and plotting can happen in parallel (2 3060s can likely plot 80TB per day, 3 can do 120 TB of plots per day, and so on). 
 * **Some farmers will actively support the new format**: I've been told by numerous community memebers that they stopped farming once compression became mainstream because of high electricity costs in their area, and that they'd return once the new plot format was out. Aside from anecdotal evidence, rational farmers might start plotting well before the new format is favored in order to 'bet' that, in the near future (when the transition period is over or almost over), new plots will have better expected return as few big farms might not be replotting or might not have finished replotting. 
 * **The change will likely cause the network's Nakamoto Coefficient to recover**: The grind-ability of the last plot format led to an alarming share of netspace using NoSSD. This is concerning because NoSSD centrally produces blocks, which led to Chia's Nakamoto Coefficient to go from [134 to 5](https://xch.farm/decentralization/). The new Proof of Space format is expected to have a higher Plot Stability Index (as defined in CHIP-0048) and resistance to grinding, which will mean all plot formats will offer similar benefits. Given this choice, I think most farmers will choose to make their own blocks, finally 'repairing' Chia's Nakamoto Coefficient and making it retake it's rightful position as the most decentralized blockchain (as measured by NC). Farmers coming from NoSSD would also run their own full nodes, improving Chia's node count as well. This elephant in the room is, to me, the most important issue the new PoS construction would solve.

With these arguments laid out, I'd like to ask that the hard fork activates in the **second week of January**, and that **the phase-out period takes 32 days**. I consider these to be reasonable timelines assuming Chia 3.0 is out before January 1st, 2026. As is the spirit of CHIPs, I'm interested in hearing arguments for and against short timelines post-Chia 3.0.

Thank you very much for this CHIP, and for all the work put towards the new format so far. Chia's new consensus and its approach to decentralization were the main things that first attracted me to the community. I'm glad to see the vision is still alive, and I hope we can adopt a timeline that puts the 'grinding phase' behind us as soon as possible now that there's a clear path forward.

Acer854:
I fully support Yak's post. I don't have much more to say other than the security and robustness of the chain outweighs trying to avoid ruffling some feathers.

SlowestTimelord:
Congrats on the CHIP release! Will provide my thoughts soon.

Rigidity:
I'm still looking everything over, but will also have some feedback soon. Excited for NC to increase! 🎉

hoffmang9:
> It is my belief that there should be no downside to replotting aside from losing compression. New plots should be 100% effective from the start if it is possible to do so. Otherwise there is a deterrent toward replotting. As a small farmer, there shouldn't be a question between do I plot for an hour a day replacing plots, wait 'till the middle and spend 2 days replotting everything, ect. The incentive should be to get done as quickly as possible.

This could have a perverse effect of making Netspace fall very hard in the first hours and days of the switch. That's why a hard cut over is dangerous from an overall security perspective.

cracklingice:
> > It is my belief that there should be no downside to replotting aside from losing compression. New plots should be 100% effective from the start if it is possible to do so. Otherwise there is a deterrent toward replotting. As a small farmer, there shouldn't be a question between do I plot for an hour a day replacing plots, wait 'till the middle and spend 2 days replotting everything, ect. The incentive should be to get done as quickly as possible.
> 
> This could have a perverse effect of making Netspace fall very hard in the first hours and days of the switch. That's why a hard cut over is dangerous from an overall security perspective.

While I agree that there is a potential for this - I do not see that being the likely outcome.  You pose a risk that farmers will shut off entire farms while I do not believe that would be the case as there is still an incentive to farm old plots so most farmers will remove and replace at the smallest level possible for their setup.  Be that plot by plot (I personally wrote a batch file that replaced old plots one at a time when I replotted for pooling and compression) or disk by disk.  I just don't foresee a situation where anyone takes down their entire farm for any reason other than they choose not to replot at all.
I would also re-state that the losing compression downfall is already a built in incentive to spread out by some amount as a fully min/max setup would see replacing all plots around the same time old plots lost your compression percentage in effectiveness.
I also would say that there is no hard cutoff in what I suggest.  There is just no additional punishment for replotting early aside from losing the amount of compression as I am suggesting that it be 100% new and 100% old at the start with each block being capable of being formed from either.  Then the first stage could be every 10th block only accepts new plot proofs (10% reduced effectiveness), ect.  This could be for example every block height ending in '0'.  Then every block height ending in '0' or '5' for 20% reduced effectiveness.  That means that people who are 40% higher than actual capacity would probably not even start replacing plots until at least the 10%, 20%, 30% or even 40% reduced effectiveness epoch depending on how fast they can plot.  There would also likely be some people that plot out space prior to the switch.  Perhaps that even causes an increase in netspace rather than decrease.  There are a lot of factors at play here.
Ultimately, I just feel that the reduced effectiveness of new plots is a punishment to the people trying to do what the is the ultimate end goal that only serves to complicate something needlessly.

hoffmang9:
I don't think we want to underestimate the lure of `rm -rf /plots/*`...

cracklingice:
> I don't think we want to underestimate the lure of `rm -rf /plots/*`...

If the user was going to do this, they will do this anyway.  All the increasing effectiveness of new plots does is make the calculation of when more difficult and disincentivizes the only actions that would defend against it (having more new plots on the network at 100% already).  In your situation - where the new plots are less effective - if we hit that 30 to 40 percent reduced effectiveness mark and new plots are only 30 to 40 percent effective instead of 100% - what happens if there is a widespread fall off of solo/official pool plots at the same time?  The people who have replotted don't provide efficient security so we could end up in a situation of disproportionate compressed control.  We both know that far too many of those are using unofficial means to farm that is detrimental to the network.  I will admit that I may not be seeing the forest thru the trees here, but I believe there is more risk with the additional disincentive toward replotting than if it was not there.

hoffmang9:
I suggest thinking about the 50%-50% point as the important time. The game will be that to maximize earnings the largest farms with the most GPUs will want to wait until just before that point to start their replot. Smaller farms will probably want to replot before them as a play to have maximal earnings against any larger farm that waited a bit too long.

Also starting at 100% effectiveness for new means a nasty game of chicken that I expect a lot of farmers would resent.

cracklingice:
> I suggest thinking about the 50%-50% point as the important time. The game will be that to maximize earnings the largest farms with the most GPUs will want to wait until just before that point to start their replot. Smaller farms will probably want to replot before them as a play to have maximal earnings against any larger farm that waited a bit too long.
> 
> Also starting at 100% effectiveness for new means a nasty game of chicken that I expect a lot of farmers would resent.

100% effectiveness of new plots is already their compression percent less effective than them in the first epoch.  In fact most current farmers would not even reach 1:1 parity with new plots in the hypothetical version I proposed until the 2nd thru 4th reduction epoch.  Anyone that would resent that has unreasonable expectations unless the reduction epochs are set wildly too close together.  Epochs don't even need to really be the same amount of time.  It could be for example that the first 30% takes 6 weeks while the remainder is cut down to 1 week per 10% leaving it roughly still half way thru that the rough 50/50 point is reached.  
Or I suppose if a new whale entered the fray with 100% new plots ready for day one that could cause some resentment, but that is far less financially reasonable than a miner who already has plots doing it.
Ultimately, I've expressed my view.  I appreciate that you took the time to reply to it even if you disagree with it.
Honestly, I'm not really all that hurt either direction as I have a NAS with free space that can hold a large majority of my current plots so I can increase my effectiveness during the transition rather than lose it.  My primary concerns are that of security and of simplicity for farmers.  Who knows - maybe I'll help the simplicity part by making a spreadsheet to help others plan.

maldivesguide:
**I disagree with the commenters asking to shrink the transition window.**

- Linear decay of existing plots over 256 days seems sensible. It allows large farmers to balance plotting time needed vs existing compression, with a clear incentive to switch. Small farms will naturally be more agile.
- Remember that the new format is intentionally slow to plot (~40 TiB per day on a 3060).

**I'd argue that 12 months is too long for the new format to start.** 

The most compressed plots are in the 200%-300% range, so the incentive to replot these plots is around +4 months from the new format going live, or 16 months from now. It's these plots specifically which have driven NC down. Can't we bring the effective date forwards by six months?

**Plot difficulty schedule doesn't make sense**

A farmer plotting minimum difficulty would need a full replot every 2 years. Alternatively, by spending ~60% more time plotting +1 difficulty the replot is reduce to every 4 years. From a farming perspective, it doesn't ever make sense to plot the network minimum. Unless we need this 2 year cadence and granularity for griding resistance the minimum difficult should be +2 every 4 years to avoid unnecessary and inefficient replots.

reythia:
**I disagree with the commenters asking to shrink the transition window.**

Linear decay of existing plots over 256 days seems sensible. It allows large farmers to balance plotting time needed vs existing compression, with a clear incentive to switch. Small farms will naturally be more agile.
Remember that the new format is intentionally slow to plot (~40 TiB per day on a 3060).

**I'd argue that 12 months is too long for the new format to start.**

The most compressed plots are in the 200%-300% range, so the incentive to replot these plots is around +4 months from the new format going live, or 16 months from now. It's these plots specifically which have driven NC down. Can't we bring the effective date forwards by six months?

**Plot difficulty schedule doesn't make sense**

A farmer plotting minimum difficulty would need a full replot every 2 years. Alternatively, by spending ~60% more time plotting +1 difficulty the replot is reduce to every 4 years. From a farming perspective, it doesn't ever make sense to plot the network minimum. Unless we need this 2 year cadence and granularity for griding resistance the minimum difficult should be +2 every 4 years to avoid unnecessary and inefficient replots.

reythia:
@Yakuhito 
> I'd like to ask that the hard fork activates in the second week of January, and that the phase-out period takes 32 days

@SlowestTimelord 
> My initial thought is a phase-out period of 64 days with a phase-out counter starting with a 25/75 split to new/old format respectively.

@Ganbin 
> Phase-out phase : 90 days

Curious to challenge all of you directly on this - you're all respected members of the community with a deep interest and commitment to the Chia ecosystem. Maybe I'm not seeing what you are.

What are you actually expecting to achieve with a shorter transition?

Bryan-J792:
in my opinion, new plot format should be activated much sooner after 3.0.  Say 30 days to activate the hardfork.  At that time, both plots contribute equally.  after 3 months the old plot phase out begins.  3-6 months later old format is completely deprecated.

madMAx43v3r:
> With these arguments laid out, I'd like to ask that the hard fork activates in the **second week of January**, and that **the phase-out period takes 32 days**.

I've re-plotted a 16PB farm before, using overkill hardware (20TB RAM + 20x RTX 4000). It took almost 2 months, but the planning and setup before took quite a while as well. Farmers will need to dust off their plotters, connect them all up again, debug a lot of issues again, etc etc. Anything below 6 months will have you lose netspace from large farms.

cracklingice:
> > With these arguments laid out, I'd like to ask that the hard fork activates in the **second week of January**, and that **the phase-out period takes 32 days**.
> 
> I've re-plotted a 16PB farm before, using overkill hardware (20TB RAM + 20x RTX 4000). It took almost 2 months, but the planning and setup before took quite a while as well. Farmers will need to dust off their plotters, connect them all up again, debug a lot of issues again, etc etc. Anything below 6 months will have you lose netspace from large farms.

As I have no experience at this level, I was wondering what your opinion on new plot effectiveness would be.  Would you prefer the new plots have reduced effectiveness or be fully effective?  With a farm of your size what would make you consider deleting all plots to start the process rather than replacing in a more granular fashion such as plot by plot, disk by disk or controller by controller?  And would there be a difference in this plan if new plots were not being limited in effectiveness?
It sounds like a reasonable timeframe would be around 1 month per 10% reduction epoch.  This would provide a full 10 months of phase out, but IMO it is critical that new plots are not limited; as the more compressed plots that are replaced by good faith farmers, the less good faith security we have to counteract any potential bad faith actors.  Or at least that new plots ramp up in effectiveness on a much faster curve than the decay.  Perhaps 3 months vs the 10.  Anything to reduce the bath tub curve that likely will favor the highest compression levels if both plots are 50% effective at the same time.

ugr-dual:
Comment on the "a 3060 can do 40TB a day" statement.

That output is a theoretical number and will be much lower due to drive speeds, usb, network, controller load, downtime, etc. - farms are very different. A pro has better tools with flash buffer space and/or fast networking. A kind of regular farm below 1-2PB will probably do the rm -f for 2,3,4 drives, fill them again and add back to the farm.

I think the transitioning perid should be as painless as possible. 90 days is very short, 6 months sounds more practical to me.

Bryan-J792:
I think we also need to state the overall goal of the transition. is it to prevent NOSSD from gaining 51% of the network?  A drop in netspace with a high NC doesn't seem like it would be an issue without a possible 51% attack.  If thats the case, we could temporarily lower the plot filter to disrupt those with high compute

danieljperry:
> I think we also need to state the overall goal of the transition. is it to prevent NOSSD from gaining 51% of the network? A drop in netspace with a high NC doesn't seem like it would be an issue without a possible 51% attack. If thats the case, we could temporarily lower the plot filter to disrupt those with high compute

The goal is simply to give everyone ample time to replot their farms without causing any major disruptions. It's not about singling any individuals out or intentionally disrupting those with high compute.

Bryan-J792:
> > I think we also need to state the overall goal of the transition. is it to prevent NOSSD from gaining 51% of the network? A drop in netspace with a high NC doesn't seem like it would be an issue without a possible 51% attack. If thats the case, we could temporarily lower the plot filter to disrupt those with high compute
> 
> The goal is simply to give everyone ample time to replot their farms without causing any major disruptions. It's not about singling any individuals out or intentionally disrupting those with high compute.

What major disruptions could be caused by a drop in netspace?

Bryan-J792:
I"m willing to bet most farmers can replot their entire farm in 1-2 months.  As farms grow, so do plotters.  

Maybe Gene or someone with a large chia following could run a poll on reddit and X.  Then it would just be what others have said.  how long does it take for people to transition to the upgrade and based off the poll, how long to replot.

Bryan-J792:
> I"m willing to bet most farmers can replot their entire farm in 1-2 months. As farms grow, so do plotters.
> 
> Maybe Gene or someone with a large chia following could run a poll on reddit and X. Then it would just be what others have said. how long does it take for people to transition to the upgrade and based off the poll, how long to replot.

would also be interesting to put up a "I don't plan to replot" option too

madMAx43v3r:
> I was wondering what your opinion on new plot effectiveness would be. Would you prefer the new plots have reduced effectiveness or be fully effective.

I would prefer they have 1:1 effectiveness from the start if I was a larger farmer. Meaning effective and physical space being the same. Since you lose the old compression on the replot.

But from older discussions I remember giving the new plots a boost, not a nerf? To avoid this individual netspace loss at the beginning of the re-plotting phase.

reythia:
> I"m willing to bet most farmers can replot their entire farm in 1-2 months. As farms grow, so do plotters.

Disagree.

Something I think many people miss is that whilst large farms may have a bit more GPU power, it's already in use for farming. HDDs and GPU are not even necessarily colocated, and many slow-plotted these large farms by expanding gradually. Plotting at scale has a lot of challenges that might not be encountered by small farms.

At anything like the phase-out times some people are calling for I'd fail to replot everything in time and a chunk of space would drop out of the network. I expect that would happen to a majority of large farms.

cracklingice:
> > I was wondering what your opinion on new plot effectiveness would be. Would you prefer the new plots have reduced effectiveness or be fully effective.
> 
> I would prefer they have 1:1 effectiveness from the start if I was a larger farmer. Meaning effective and physical space being the same. Since you lose the old compression on the replot.
>
> But from older discussions I remember giving the new plots a boost, not a nerf? To avoid this individual netspace loss at the beginning of the re-plotting phase.
> 
 "The phase-out period would then take an additional 256 days, where the 'weight' of old plots lowers while that of plots using the new format increases." makes it sound like the way that I had understood it where new plots would start at very low effectiveness and increase in effectiveness inversely to the decrease of the old plots.  Basically - creating a bell curve where somewhere around 40% or 30% effectiveness on old plots we're likely around the bottom of a bell curve where I believe we would see network capacity around 50% of current.
 >
> > With a farm of your size what would make you consider deleting all plots to start the process
> 
> You never do that, always delete in batches, like per JBOD or something.
> 
This is exactly what I thought would be the case but Gene seems to fear a mass rm -rf *.plot party or something - and I can somewhat see this when folks finally start ditching nossd and any other entirely incompatible plot formats - although I still think anyone that absolutely cannot bring different farmers online would likely remove drive by drive to replot and only 'mass delete' once they're about halfway thru effective capacity and can bring back the harvester by swapping to just the drives using new plots.
Also - I fully anticipate a withdrawal of at least 5 to 10% of raw netspace that simply won't come back due to burnout, but I could also see a small potential where the change ends up increasing profits and causing a surge of newcomers.

BrandtH22:
One farmer type I am not seeing discussed here are set and forget, low technical knowledge farmers. Generally these farmers are small to mid size and have mostly self-contained farming rigs but these are definitely the types of farmers that would rm -rf plots once it makes sense for them to do so (i.e. they want to remove the plots, start the plotting of new, and not touch it again until the next update).

Another aspect that has not been discussed much is plot tools that the community can help to build which would provide a more seamless upgrade path for these types of farmers and perform the plot replacements on their behalf.

Personally on my farm I am likely to either use a tool like the above or just nuke plots a few drives at a time.

BrandtH22:
As for activation timeline, I believe (based on the chia dashboard version history) that 30 days is far too short and the shortest time period that should be considered is 60 days but 90 at a minimum would be better. It is critical to have a majority of nodes/space transitioned before that activation and as mentioned there are a large number of set and forget farmers.

danieljperry:
The term `plot difficulty` could be confused with the network's global `block difficulty`. In addition, while "difficulty" in this context describes the amount of time required to create a _plot_, it could easily be confused for the time required to create/solve a _proof_. With these considerations in mind, we have decided to update the term `plot difficulty` to `plot strength`.

danieljperry:
This CHIP is now in `Review` status. Please leave your reviews here, and feel free to discuss it in the #chips channel of [our Discord](https://discord.gg/chia).

PR #93, opened 2023-12-12
Introduce DAO CHIP

danieljperry:
Adding a new CHIP for a DAO standard.

PR #165, opened 2025-07-23
CHIP-0050 & CHIP-0051: Action Layer and Slots + Reward Distributor

Yakuhito:
> [!NOTE]
> **Low Risk**
> Documentation-only additions to the CHIPs repo with no executable or protocol changes in this PR.
> 
> **Overview**
> Adds two new **Final** informational puzzle CHIPs to the repository.
> 
> **CHIP-0050 (Action Layer and Slots)** documents an optional singleton inner-layer framework for Chia dApps: Merkle-tree–gated **actions** runnable in one spend, **ephemeral** vs **persistent** state, pluggable **finalizers** (default and reserve), and **slots** for coinset-backed persistent data outside the main state.
> 
> **CHIP-0051 (Reward Distributor)** specifies an on-chain reward distributor built on CHIP-0050 (and CHIP-0035 for curated NFT mode): per-second weighted distribution, epoch transitions, sponsor **commit/withdraw** with penalties, and launch-time modes (**managed**, collection NFT, curated NFT, CAT) with the corresponding action sets (`sync`, `new_epoch`, stake/unstake, payouts, etc.). Both CHIPs point to reference puzzles, `chia-wallet-sdk` tests, and security notes; they do not change protocol or mandate adoption.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 19cebd83e28fdb7b2e560ab63837114c6c332615. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


danieljperry:
CHIPs 50 and 51 are now drafts. (Both CHIPs can be handled in this PR.) Please leave your reviews here, and feel free to discuss in the #chips channel of [our Discord](https://discord.gg/chia).

Yakuhito:
> Left a comment PR on the reference repository: [Yakuhito/slot-machine#13](https://github.com/Yakuhito/slot-machine/pull/13)

[Replied](https://github.com/Yakuhito/slot-machine/pull/13#issuecomment-3124469298). The slot puzzle no longer assumes a singleton amount of 1 and is roughly ~25% smaller.

Yakuhito:
After thinking about your ideas @OverActiveBladderSystem here's my comments:

> is these concepts something that could be used for farming on pools (or solos) when it comes to claiming and distributing rewards with fewer transactions, as well as flexibility in how much of the initial reward portions are committed to the pool and/or farmer?

> Fantasy 1: would this allow a fancy puzzle for farmers to commit 100% (or varying amounts) of the pool/farm reward portions to a pool rather than the 87.5% / 12.5% split today?

To change the split, you'd probably want to change consensus to allow custom or even configurable percentages. I think the same mechanism used to switch pools could work. I'm not an expert on that topic, so not the best person to ask 😅

> If transaction fees in the farmer portion become the largest portion of the reward, some users will want that shared with the whole pool. So if there was an "official protocol" instead of a centralized pool making blocks for them or trusting people won't cheat a regular pool which often requires them to build up credit owed before payout, that might be neat.

This would, more likely, require a custom pooling singleton (and that to be supported at the consensus level - not sure if it is). I'm pro-official-protocol-pools, but I don't think these puzzles can necessarily help here.

> Fantasy 2: would this allow stacking of multiple pool rewards and a pool can inject their (by block, by day, by fixed amount) oracle data for payouts when claiming those rewards, turning it into a single transaction? rather than claiming rewards and multiple distribution transactions after?

The distributors work on a per-second basis. The difficult 'balance' here is that pools will likely frequently update the list of active farmers and their shares of the rewards, which might also cause a lot of on-chain transactions. This contrasts the ideal case of the rewards distributor, where updates from the manager are not that frequent. The [NFT reward distributor](https://blog.fireacademy.io/i/166151150/what-dig-is-doing) doesn't have this problem because users update it themselves when they stake/unstake an NFT, but you unfortunately can't trust farmers to update their own space. Is there an aspect of the current farming rewards distribution mechanism you'd like to see improved? 

> If the fantasy thinking is possible, would it just take a change to consensus in the upcoming hard fork to open the door to these ideas and more with flexibility going forward while we're doing the hard fork?

Generally no, as the CHIPs cover Chialisp puzzles that can run on the current version of the blockchain. I think a hard fork would be needed for the split idea, for example. Probably best to ask over Discord, where people with better knowledge in that area can answer.

Yakuhito:
[CHIP call slides](https://pitch.com/v/chip-0050-chip-0051-xymjnx) (call in 8h13m from now)

Yakuhito:
The NFT mode of the reward distributor has begun mainnet testing as part of the DIG Network community alpha: [X post](https://x.com/digdotnet/status/1961449065574441198) / [site](https://rewards.fireacademy.io/)

Yakuhito:
Hey, all! As I mentioned in a DM I sent @judeallred shortly after his comment, I held off responding to the last comment until the new version of the reward distributor was tested and ready to be discussed. I'll address the points below, then send another (independent) comment detailing the update. Please feel free to ask for clarifications/post new questions that may arise.

> After claiming, I received 75 DIG coins, each of amount 0.016, and there are more to claim. Claiming a second time...
I received 31 more coins, also of 0.016 DIG each.
> 
> It appears the rewards batch per NFT. In the single-NFT case this is good, as I receive one reward coin per claim. In the multi-NFT case, it'd be nicer if I received one coin overall. At present, following a claim, I need to do a sweep operation in my wallet in order to keep the rewards manageable.
> Would it be possible to bundle the rewards into one coin, no matter how many NFTs are staked?
> Is it viable to adjust the claim mechanism to gather all of the rewards in one operation (instead of one-per-75)?

For the first question: yes! The reward distributor was initially designed for DIG, and the NFT mechanism was added as an extension. DIG mirrors would only have one entry ('ticket'), while, in the first version of the NFT reward distributor, each NFT acted as its own, independent entry. I severely underestimated the number of people holding 10+ NFTs of a given collection. With the new version, multiple NFTs (or CATs, more on that later) can be consolidated into the same entry - there will be one 'ticket' per user, not per NFT. This means you'll be able to get your payout in one coin only, with no real disadvantages other than slightly more complex staking/unstaking code.

For the second question: limits were set so spend bundles do not reach the maximum cost allowed for mempool items (5.5 billion - half a block). Now that payouts are done in a single coin for all your NFTs, you won't come near that limit - so you'll get all your rewards in one claim. A lot of users have asked for this, and I'm glad the new version of the reward distributor will have a much better experience. The only operation where you'll still have to do 'batches' is staking and unstaking - unfortunately, moving NFTs around is really expensive, so I don't think you'll be able to stake (or even transfer) 103 NFTs in a single block. Still, I think work on the drivers/frontend could improve the experience on that front (e.g., upload 5 offers one after the other, and the site automatically tries to stake one batch per block until the queue is empty).


> e.g. It'd be neat if Tang Gang could contribute HOA to a given DataLayer epoch if they wanted to. DataLayer would need to 'approve' this (or add the HOA on themselves) the so that an epoch's rewards don't get dusted or include un-desired cross-promotions.

For the purpose of this CHIP, only one reward token is allowed per reward distributor. While it's possible to have more than one token, it would require a big architectural change. If the demand is there, a future reward distributor version could be built to allow this along with a new CHIP.

A lot of distributors may work with just _one_ reward token, and assuming one reward token only allowed me to make a lot of optimizations in the current code. It may make sense to still have them run with the current standard, and then anyone who wants to take sponsorships in another token could use the other standard (which would be more costly given the increased complexity). Much like it makes sense to separate asset-asset partial offers from multi-asset ones, I think this distinction is important here as well. 

> The ability to stake CATs as well as NFTs. Perhaps your spec already permits this, but I would say broadly that in other ecosystems it's usually tokens that are staked rather than NFTs. This is broadly used to drive up the price of a token in the short term (/fundraise) in exchange for promising an APR on that token once staked, with the hopes that the token's project will in time become more valuable as the project succeeds.... or it's used as a sort of pump-and-dump by the creator. That said, lots of non-scam popular projects have this and used it to bootstrap their funding / reward early investors.

To quote AI, "You're absolutely right!" - the new version of the reward distributor has a mode where you can stake CATs to earn rewards. The main use case I see for this type is allowing projects to incentivize TibetSwap liquidity, but I'm sure the community will come up with entirely new ways to use this primitive.

> Flexibility in staking durations. I've seen a few common patterns here (...)

Really appreciate the examples you're providing! All the modes included in this new CHIP will allow the users to unstake assets at any time. That said, at the Chialisp level, there is an easy way to add these kinds of timed locks via custom locking/unlocking puzzles. In short, you'd only have to write the locking/unlocking puzzles for a distributor that locked tokens according to the rules you mentioned above - you could then reuse most of the code of the current distributors (slot consolidations, payouts, accounting, etc.).

> Staking and rewards coupled into a DAO mechanism, where a staked position (including relevant multipliers) grants voting power in proportion to the stake and its multiplier, and epoch rewards can be scaled depending on the user's level of participation in the DAO's votes (e.g. Helium, Sanctum, and Jupiter, above)

DAOs on Chia are another beast (see [CHIP-0024](https://github.com/Chia-Network/chips/pull/93)), and definitely a major undertaking. I think reward distributors could remain mostly independent and make this scenario possible by having the DAO mechanism designed around them. It seems it would only require custom locking/unlocking puzzles. Each action already produces announcements - right now, users use those to secure the bundle, but dApps may as well use them to assert certain operation were done on the distributor.

> Liquid staking tokens.

These are actually very interesting! The fact that the entries become fungible (sTokens instead of users each having a 'ticket' to reclaim staked assets) means the reward distributor is not ideal for them, but you could use something extremely similar, as the reward 'accumulation' mechanism (which I wrote about [here](https://blog.fireacademy.io/p/reward-distributors-101)) is the same. Honestly, this sounds like a really exciting project from a technology point of view.

> I imagine many of these would be large efforts deserving separate CHIPs, but since you've thought through the Chialisp on this, perhaps some of these items are viable to bring into scope for this CHIP.

Once again, thank you very much for the interesting ideas! Your comment heavily influenced the decision to generalize locking/unlocking into puzzles to keep reward distributor flexible so that future extensions will require minimal work (specifically timed locks). While some features are addressed by the new version (slot consolidation/CAT reward distributors), the line between this CHIP and future ones needs to be drawn somewhere. The reward distributor in particular is a very complex dApp where every new feature needs careful design, optimization, testing, and reviews - while all the cases you mentioned are very interesting, I think we should be very wary with scope creep, as any new feature would require a lot of effort. That said, I think this next deployment will show the capabilities of the general system on Chia, and I hope it will inspire everyone to keep building on top of this CHIP.

Yakuhito:
A recent security-focused review found that an update aiming to prevent singletons with the action layer (CHIP-50) from being spent while no actions are run did not achieve its purpose. The [new code change](github.com/Yakuhito/slot-machine/commit/ddd99fc213325babfa9e1b915b9372600c8f7729) removes the old fix and adds a correct one.

Lowestofttim:
Do we have any updates on this chip?

danieljperry:
> Do we have any updates on this chip?

@Yakuhito

Yakuhito:
Given the complexity of the code, we're waiting for one more review before proceeding to the new Alpha release. I'll reach out to interested project once everything is in place.

Lowestofttim:
Thanks Yak, I don't mean to rush you 🫡

Yakuhito:
Hey, everyone! Two more reviews have taken place since the last update, and we're in a place where we're confident in the security of the (new) hardened puzzles. You'll notice an additional warning in the action layer CHIP - slot value re-use is a new type of attack that developers need to be aware of. For example, [this version](https://github.com/Yakuhito/slot-machine/blob/312c423d617438f0b09e1a341b890435f4b029af/puzzles/actions/xchandles/oracle.clsp) of the XCHandles oracle action could be 'tricked' into verifying arbitrary information. All slots where it's not trivial to prove a re-use attack is impossible now include incrementing counters in their values as a defensive measure.

[slot-machine](https://github.com/Yakuhito/slot-machine) and [rewards-distributor-clsp](https://github.com/DIG-Network/reward-distributor-clsp) have been updated with the latest code. The former contains a CLI that can be used to interact with the latest version of the primitives in this CHIP (as well as XCHandles). The driver code is being integrated into chia-wallet-sdk through [this PR](https://github.com/xch-dev/chia-wallet-sdk/pull/425).

With these updates, we're also refreshing the [rewards.fireacademy.io](https://rewards.fireacademy.io/) site. When we're ready for the next community alpha, I will reach out to participants that expressed interest (as always, contact me if interested). The new release is intended to help us refine the UX/experience around distributors rather than the underlying (on-chain) logic.

As the puzzles have reached the final intended form, this CHIP will progress towards being finalized.

danieljperry:
CHIPs 50 and 50 are now in `Last Call`. If no additional changes are required in the next two weeks, then they will become `Final`.

danieljperry:
CHIPs 50 and 51 are now `Final`. No further changes are allowed, other than adding errata.

danieljperry (approved):

Rigidity (approved):

PR #197, opened 2026-03-05
Create chip-Jake15x0-Alternate Farming Terminology Modes

jake15x0:
Create chip-Jake15x0-Alternate Farming Terminology Modes

### DIFF
--- a/CHIPs/chip-Jake15x0-Alternate Farming Terminology Modes
+++ b/CHIPs/chip-Jake15x0-Alternate Farming Terminology Modes
@@ -1,0 +1,193 @@
+CHIP Number   | < Creator must leave this blank. Editor will assign a number.>
+:-------------|:----
+Title         | Alternate Farming Terminology Modes
+Description   | Add an advanced setting in the GUI and CLI in the PoS2.0 plotting/farming software to toggle between Tilth and Rigidity based nomenclature.
+Author        | jake15x0 (github), Jake/ZTOR (discord)
+Editor        | < Creator must leave this blank. Editor will be assigned.>
+Comments-URI  | < Creator must leave this blank. Editor will assign a URI.>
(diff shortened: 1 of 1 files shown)


danieljperry:
I also [commented](https://discord.com/channels/1034523881404370984/1041838182402117672/1479247914646765630) in our Discord, but I'll add to that comment here.

This feels like a feature request, and not a CHIP.

For this to be a CHIP, it would need to propose a standard methodology for any farming tool in the ecosystem to use, irrespective of implementation details. This instead proposes a specific feature that only the reference wallet would use, with a specific implementation.

As for my personal opinion, I'm not totally opposed to this. It could be a fun little Easter egg in the reference wallet, much like "Pirate English" currently is. But nevertheless, it's a feature, not a CHIP.

jake15x0:
I don't see this change relevant for the active reference wallet software "today".  I had CHIPs 48/49 as pre-reqs.  My proposal was for the new tooling for PoS2.0.  And yes this is more of a fun Easter egg like Pirate English or like themes in Sage Wallet which don't affect the function of the software.  
The Chia discord community was very active during the Plot Strength vote (not nearly as much now).  There is interest in this terminology and I imagine there will be jokes had, memes made, and fun by all during the replot if CNI put this in place.  I bet a main net anniversary celebration announcement that included this farming terminology would generate conversation.  At this point any conversation beyond "price bad" on discord would be good for the community.
For the exceptional developers at CNI my understanding is that something like this is easily done, however I wouldn't request anything that causes delays to the current timeline.
The definition of CHIP is broad as I read it. If it's a CHIP or a feature request I don't care.

danieljperry:
I'm not saying that I disagree with your idea. I'm saying that it shouldn't be a CHIP.

PR #202, opened 2026-05-18
CHIP-0058: Parallel voting at scale: off-chain proofs, on-chain finality

MichaelTaylor3d:
This CHIP defines a permissionless standard for on-chain voting at scale on Chia. It is practical at around 20,000 voters per election, with no known theoretical upper bound, and it solves the singleton parallelism problem that has, so far, forced every serious on-chain vote to either crawl one update per block through a Ballot Box Singleton or fall back to a semi-permissioned aggregator key. 

The key idea is that a fixed Groth16 circuit produces a zero-knowledge proof that today's CLVM can verify natively, using the BLS pairing operators already shipped in [CHIP-0011](https://github.com/Chia-Network/chips/blob/main/CHIPs/chip-0011.md). No soft fork or new opcodes are needed. On-chain verification stays constant in size, no matter how many voters contributed off-chain, and the proof binds three things at once that BLS aggregation alone cannot: the BLS signature aggregate, the quorum threshold, and the registered-voter scope. As a result, no aggregator can produce a passing finalize without the real votes behind it. Registration and ballot issuance stay on the Election Singleton's slow lane. Per-voter writes run in parallel on independent Voting Coins. Per-ballot finalize lives on the Ballot Coin. 

A separate Ceremony Singleton lineage anchors each election to a public, auditable Groth16 trusted setup at deploy. All puzzles are written in [Rue](https://github.com/Rigidity/rue), the high-level language for CLVM. The same architecture supports DIG Network's L2-style anchoring, CAT-weighted DAO governance, and Chia Vault spend paths that take authorization from a finalized ballot announcement, which makes large-scale trust-minimized governance realistic on the base layer.


danieljperry:
This CHIP has been assigned CHIP-58 and is now a `Draft`. Feel free to leave your reviews here, and you can discuss the CHIP in the #chips channel of [our Discord](https://discord.gg/chia).

MichaelTaylor3d:
The nextjs app in this repo that demos the functionality as a dapp is still waiting to have signMessageUnsafe exposed from Sage through Wallet Connect. I'm working with @Rigidity to add this endpoint.

chip0002_signMessage adds extra bytes to the message before signing (each voter's signature ends up tied to their specific public key), which means when we try to combine N voters' signatures into one for the on-chain check, they don't merge cleanly, we'd have to verify each voter individually, making the proof N times more expensive.  

the voting protocol is built around sign_raw (the unaugmented version) precisely because those signatures do combine cleanly into a single signature you can verify in one shot, which keeps the clvm cost flat regardless of voter count

danieljperry:
We are going to Discuss this CHIP in a Zoom call on June 2 at
8 AM PDT
3 PM UTC
11 PM China

See the #chips channel of our Discord for details.

danieljperry:
We need to move the call to discuss this CHIP to Thursday, same time. Sorry for any inconvenience.

danieljperry:
The recording of the discussion of this CHIP is now available on Youtube.
https://youtu.be/JeHI4Gauy7w

danieljperry:

MichaelTaylor3d:
I have identified that I'm missing a constraint in the growth circuit that can allow forged votes under specific circumstances. I'm working on a patch now.

PR #194, opened 2026-02-26
CHIP-0056 Fee CATs

cameroncooper:
This CHIP adds Fee Layer support to CATs for CAT tokens to pay transaction fees to issuers.

---

> [!NOTE]
> **Low Risk**
> Documentation-only changes to the CHIPs repo; no runtime or consensus code is modified in this PR.
> 
> **Overview**
> Adds **CHIP-0056 (Fee CATs)** as a new Standards Track / Primitive proposal in **Review**, and lists it under **Draft** in `README.md` (linked to PR #194).
> 
> The CHIP defines a **fee puzzle layer** for CAT2 (optionally stacked with CHIP-0038 revocation) that charges issuer-configured **percentage + minimum** transfer fees on **offer lock-leg trades**, enforced via **settlement payment puzzle announcements** rather than deducting from CAT ring balances. It specifies immutable issuance-time parameters, the `SetCatTradeContext` / trade-price descriptor format, spend paths (normal enforcement, settlement unlock bypass, TAIL mint/melt, optional revocation bypass), on-chain **quote settlement puzzle hash reconstruction** for CAT-quoted trades, wallet/offer integration guidance, reference implementation pointers, and security notes (layer stripping, self-announcement, descriptor tampering, under-declared prices).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2e5c7ef9824c76da06e0a8dbdaa54b3ba5f6c17a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #205, opened 2026-05-25
CHIP-0060: Titled CATs

EvanWinget:
This CHIP adds the Titled CAT Standard as a Draft CHIP. Titled CATs are an additive and opt-in standard to give any CAT the same immutable, DID-rooted on-chain identity an NFT has without relying on a registry. 

This CHIP introduces no new consensus rules and only one new puzzle, the Title Singleton. The `precision` field allows Titled CATs to support the full range of options from divisible fungibles like stablecoins to indivisible assets like trading cards or tickets.

### DIFF
--- a/CHIPs/chip-evanwinget-titled-cat.md
+++ b/CHIPs/chip-evanwinget-titled-cat.md
@@ -1,7 +1,7 @@
 CHIP Number   | TBD (to be assigned by Editor)
 :-------------|:----
 Title         | Titled CAT Standard
-Description   | A standard for CATs that carry immutable on-chain identity and provenance, spanning divisible tokens and indivisible editioned assets
+Description   | A standard attaching an immutable, on-chain metadata reference (name, ticker, content hashes, creator DID, and supply commitments) to a CAT.
 Author        | [Evan Winget](https://github.com/EvanWinget)
 Editor        | TBD
 Comments-URI  | TBD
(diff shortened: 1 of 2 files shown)


Quexington:
https://github.com/Chia-Network/chips/pull/205/changes#diff-655c6a996e67b11d1a1e4af1fce5b43c23c398a1922912cc54d5ffb058f759f4R48
- I think it's a bit wrong to say that CATs don't carry that on-chain identity and then explain how they have on-chain identity with their TAILs 😅.  You're about to hook into this TAIL to flesh out the on-chain identity but it is there, we just don't use it like NFTs.

https://github.com/Chia-Network/chips/pull/205/changes#diff-655c6a996e67b11d1a1e4af1fce5b43c23c398a1922912cc54d5ffb058f759f4R112
- I don't think you should make immutability a hard-coded part of the standard.  The immutability of fields can be set and checked by things that care and if it's important to something that they want to trust a field forever, they can check for this accordingly.  You make a good case for why immutability is important but I don't think it needs to be forced upon everybody and everything.

https://github.com/Chia-Network/chips/pull/205/changes#diff-655c6a996e67b11d1a1e4af1fce5b43c23c398a1922912cc54d5ffb058f759f4R165
- I think implying that this could subsume NFT1 is problematic.  The whole point of a CAT is that it's fungible so you'd need a bunch of checks on things like a p2_singleton equivalent that checks that the CAT has the right tail (which requires supplying the singleton ID) and that the title has the required edition cap and that the coin amount is 1 instead of 0 etc.

https://github.com/Chia-Network/chips/pull/205/changes#diff-655c6a996e67b11d1a1e4af1fce5b43c23c398a1922912cc54d5ffb058f759f4R305
- A long standing wallet struggle is that it's somewhat impossible to know what a TAIL is.  You have to follow the CATs entire lineage back to the TAIL spend to get the reveal.  You could potentially have the wallet try to subscribe to the TAIL hash which the title singleton hints to? But it's a bit of a sell to say that CATs should always make a subscription by default.

https://github.com/Chia-Network/chips/pull/205/changes#diff-655c6a996e67b11d1a1e4af1fce5b43c23c398a1922912cc54d5ffb058f759f4R309
- The problem with optional precison-ality is that you create unsafe scenarios where maybe an explorer shows a different value for an offer than the wallet does and you can kind of trick people into paying orders of magnitude less than they intended to.  Having a default in the wallet was a mistake on Chia's part for the original CAT standard but the reason it's been so sticky is the fact that it's important that everyone agrees what they're looking at.

(broad comments)
- This CHIP is quite long and repetitive and I would appreciate if it was more focused on what it was trying to accomplish which is to provide metadata references for CATs.  You give a really comprehensive of the existing landscape of Chia applications but I think it distracts from the core idea and for me at least I find it confusing and fatiguing.
- This doesn't necessarily affect the construction or necessity of the CHIP, but it's worth pointing out that NFTs have some shortcomings that make them seem more assymentrical to CATs than is strictly necessary.  Because we developed singletons primarily with pooling in mind, the design is sort of laser focused on only allowing 1 of a thing (hence the name "singleton") but in reality, that's sort of unnecessary for most applications of NFTs.  In the future, it should be possible to have NFTs where you can mint N things with the same asset ID but which must always remain N coins.  This, in particular, seems like it would be a nice use case for the trading cards that you bring up a couple of times.  And in that scenario, even NFTs may benefit from a metadata reference like the title standard here. My point is that not all problems explicitly lie with shortcomings of CATs, some of it is due to shortcomings of NFTs and how those shortcomings have been handled by the ecosystem.  Important to think about from a philosophical perspective I think.

EvanWinget:
Thank you for the review and feedback, @Quexington!

> https://github.com/Chia-Network/chips/pull/205/changes#diff-655c6a996e67b11d1a1e4af1fce5b43c23c398a1922912cc54d5ffb058f759f4R48
> 
> * I think it's a bit wrong to say that CATs don't carry that on-chain identity and then explain how they have on-chain identity with their TAILs 😅.  You're about to hook into this TAIL to flesh out the on-chain identity but it is there, we just don't use it like NFTs.

This is correct and I was imprecise here. A CAT does have on-chain identity via its asset ID. What I meant is that a CAT lacks *human-meaningful* metadata and provenance bound to that identifier. I have modified the CHIP to be more precise here.

> https://github.com/Chia-Network/chips/pull/205/changes#diff-655c6a996e67b11d1a1e4af1fce5b43c23c398a1922912cc54d5ffb058f759f4R112
> 
> * I don't think you should make immutability a hard-coded part of the standard.  The immutability of fields can be set and checked by things that care and if it's important to something that they want to trust a field forever, they can check for this accordingly.  You make a good case for why immutability is important but I don't think it needs to be forced upon everybody and everything.

My take is that for an identity-bearing standard the downsides of mutability outweigh the benefits. A malicious change to an asset's metadata after holders acquire it is a worse failure than an uncorrectable typo, and if immutability is per-field and optional, every consumer has to track "is this field frozen, who may change it, and have they?". A holder of Spacebucks should not be in a position where the issuer can change the asset out from under them IMO.

I'll also note the Titled CAT proposal is additive and opt-in, so it does not force immutability on every CAT. An issuer who wants mutable branding still has plain CAT2 plus a registry. Also if the metadata genuinely must change there could be an escape hatch via melt-and-mint migration + offer files.

If you can share reasons that mutability is valuable and what the UX might be, I'd be happy to learn more about the benefits and use cases. 

> https://github.com/Chia-Network/chips/pull/205/changes#diff-655c6a996e67b11d1a1e4af1fce5b43c23c398a1922912cc54d5ffb058f759f4R165
> 
> * I think implying that this could subsume NFT1 is problematic.  The whole point of a CAT is that it's fungible so you'd need a bunch of checks on things like a p2_singleton equivalent that checks that the CAT has the right tail (which requires supplying the singleton ID) and that the title has the required edition cap and that the coin amount is 1 instead of 0 etc.

Agreed and I have walked this back now. NFT1 remains the right option for non-fungible tokens.

> https://github.com/Chia-Network/chips/pull/205/changes#diff-655c6a996e67b11d1a1e4af1fce5b43c23c398a1922912cc54d5ffb058f759f4R305
> 
> * A long standing wallet struggle is that it's somewhat impossible to know what a TAIL is.  You have to follow the CATs entire lineage back to the TAIL spend to get the reveal.  You could potentially have the wallet try to subscribe to the TAIL hash which the title singleton hints to? But it's a bit of a sell to say that CATs should always make a subscription by default.

You're right, thanks for pointing this out. tcat-aware wallets SHOULD carry the launcher-ID hint forward on transfers too, which keeps the common aware-to-aware case free. A wallet needs to resolve asset_id to launcher_id once per title and can cache it permanently, since the cached value re-verifies against the asset ID, so it's one-time per title rather than per transfer. An index could also serve this mapping and stay trustless for the same reason. If a wallet meets a title with no hint and no index entry, it walks the lineage back to a mint to recover the reveal and then caches, or degrades to showing a generic CAT.

I rewrote the discovery section along these lines. This relies on the asset ID being a deterministic function of the launcher ID, so I checked CHIP-40: the everything_with_singleton NONCE is curried in only to let a single singleton issue multiple different CATs, and it never appears in the puzzle body, only the curry. Since each title has its own dedicated singleton, I think it's OK to pin the nonce to a constant. That keeps the asset ID deterministic from the launcher ID at no runtime cost and reuses the deployed puzzle unchanged. Please let me know if I am missing a reason it needs to vary.

> https://github.com/Chia-Network/chips/pull/205/changes#diff-655c6a996e67b11d1a1e4af1fce5b43c23c398a1922912cc54d5ffb058f759f4R309
> 
> * The problem with optional precison-ality is that you create unsafe scenarios where maybe an explorer shows a different value for an offer than the wallet does and you can kind of trick people into paying orders of magnitude less than they intended to.  Having a default in the wallet was a mistake on Chia's part for the original CAT standard but the reason it's been so sticky is the fact that it's important that everyone agrees what they're looking at.

I agree and this is the reasoning behind committing precision on-chain. Because precision is a mandatory immutable field on the title (as currently proposed), every wallet can read the title to agree on what a balance means. Disagreement can still arise in a legacy or non-Titled-aware wallet that does not read the field and falls back to its own default, and this requires some additional thought/discussion. 

> * This CHIP is quite long and repetitive and I would appreciate if it was more focused on what it was trying to accomplish which is to provide metadata references for CATs.  You give a really comprehensive of the existing landscape of Chia applications but I think it distracts from the core idea and for me at least I find it confusing and fatiguing.

Fair point, thanks. I wrote this across multiple sessions and it sprawled. I just did another pass for focus and conciseness (with Claude giving some advice on cutting redundancy), and addressed your feedback as best I could.

> * This doesn't necessarily affect the construction or necessity of the CHIP, but it's worth pointing out that NFTs have some shortcomings that make them seem more assymentrical to CATs than is strictly necessary.  Because we developed singletons primarily with pooling in mind, the design is sort of laser focused on only allowing 1 of a thing (hence the name "singleton") but in reality, that's sort of unnecessary for most applications of NFTs.  In the future, it should be possible to have NFTs where you can mint N things with the same asset ID but which must always remain N coins.  This, in particular, seems like it would be a nice use case for the trading cards that you bring up a couple of times.  And in that scenario, even NFTs may benefit from a metadata reference like the title standard here. My point is that not all problems explicitly lie with shortcomings of CATs, some of it is due to shortcomings of NFTs and how those shortcomings have been handled by the ecosystem.  Important to think about from a philosophical perspective I think.

This is a great point, and I added a note in the NFT1 section. On the N-of-N NFT shape: because it must remain N coins, holding or transferring multiple copies of the same card would be N-in / N-out, where fungible Units consolidate into one coin and transfer in roughly one. So I think the two still partition along individuation. The N-of-N shape fits cards that are individually distinct, and the fungible title fits interchangeable copies you hold and trade in bulk.

danieljperry:
This CHIP has been assigned as CHIP-60. It is currently a `Draft`. Please keep the reviews coming, and feel free to discuss it in the #chips channel of [our Discord](https://discord.gg/chia).

Quexington:
> My take is that for an identity-bearing standard the downsides of mutability outweigh the benefits. A malicious change to an asset's metadata after holders acquire it is a worse failure than an uncorrectable typo, and if immutability is per-field and optional, every consumer has to track "is this field frozen, who may change it, and have they?". A holder of Spacebucks should not be in a position where the issuer can change the asset out from under them IMO.

> I'll also note the Titled CAT proposal is additive and opt-in, so it does not force immutability on every CAT. An issuer who wants mutable branding still has plain CAT2 plus a registry. Also if the metadata genuinely must change there could be an escape hatch via melt-and-mint migration + offer files.

> If you can share reasons that mutability is valuable and what the UX might be, I'd be happy to learn more about the benefits and use cases.

It can be the recommendation for wallets to only pay attention to provably immutable stuff at the beginning.  The problem is that baking it into the standard is unnecessarily restrictive and having the response "someone can go use another standard" kind of undermines the necessity of a proposed standard.  URIs are a common example of metadata that may need to change just because links break sometimes and need to be updated.

> You're right, thanks for pointing this out. tcat-aware wallets SHOULD carry the launcher-ID hint forward on transfers too, which keeps the common aware-to-aware case free. A wallet needs to resolve asset_id to launcher_id once per title and can cache it permanently, since the cached value re-verifies against the asset ID, so it's one-time per title rather than per transfer. An index could also serve this mapping and stay trustless for the same reason. If a wallet meets a title with no hint and no index entry, it walks the lineage back to a mint to recover the reveal and then caches, or degrades to showing a generic CAT.

> I rewrote the discovery section along these lines. This relies on the asset ID being a deterministic function of the launcher ID, so I checked CHIP-40: the everything_with_singleton NONCE is curried in only to let a single singleton issue multiple different CATs, and it never appears in the puzzle body, only the curry. Since each title has its own dedicated singleton, I think it's OK to pin the nonce to a constant. That keeps the asset ID deterministic from the launcher ID at no runtime cost and reuses the deployed puzzle unchanged. Please let me know if I am missing a reason it needs to vary.

That's a good change, but I don't think I explained my point properly.  My point is that when you receive any generic CAT, you can't know if it has a title or not.  In order to check, you'll have to register interest in the TAIL hash and see if there's the appropriate title singleton returned to you.  This is potentially a lot of unnecessary wallet subscriptions and also, I didn't mention before, but it's a very juice target for dusting since many wallets will all try to subscribe to this same hint.

> This is a great point, and I added a note in the NFT1 section. On the N-of-N NFT shape: because it must remain N coins, holding or transferring multiple copies of the same card would be N-in / N-out, where fungible Units consolidate into one coin and transfer in roughly one. So I think the two still partition along individuation. The N-of-N shape fits cards that are individually distinct, and the fungible title fits interchangeable copies you hold and trade in bulk.

Still not super relevant to the CHIP but trading cards are indeed a gray area of fungibility.  Like you mention in the CHIP, once you start wanting like edition or serial numbers, then moving back to NFTs seems to be the move but using CATs otherwise might make sense.  The only thing I was trying to bring to mind is that right now the different editioned/serialed cards would have necessarily different asset IDs according to the current standard whereas in the future you might be able to have a game of some sort check for "a Pikachu NFT" where all of the Pikachu NFTs have the same asset ID just like how CATs work and without having to solve for what edition, etc. they are.  And your title standard would be helpful for centralizing the image of a Pikachu among the 1000 or whatever that you minted.

Yakuhito:
A few points from my first pass:
 * This does not work for all CATs, only those with a specific TAIL (or whose TAILs reference singleton launcher ids). Most CATs on Chia currently use the single-issuance TAIL.
 * (Small) Wouldn't `precision` taking a value that translates to the number of decimals of the token be a bit more intuitive? So `0` means whole units, and `3` (not `1000`) is the default. Might be the Chialisper part of my brain saying this would take a few less bytes, but also this gets rid of the 'must be power of 10' rule.
 * (Idea) In the CATalog CHIP, there are two versions of names/tickers/etc. - one all wallets/frontends can use (and they'll treat the CAT as having a precision of 3), and one to be used by interfaces supporting custom precision. This way, the legacy wallet would show you 123.456 milliETH, while Sage could show you 0.123456 ETH for the same asset.
 * Should royalty config or revocability live in the metadata at all? Those can be parsed from the CAT layers when you receive a CAT.
 * I (also) do not agree with immutability here. If an attacker compromises a title singleton, they can mint as many CATs as they wanted (revocation would also be a possible issue) - metadata would not be a major concern for the CAT holders, unless names change to impersonate other assets (which can be solved by CATalog's verifier layer, for example). As someone who maintains an 'immutable' list of verified CAT details, I've had many requests to update images (not just fix broken URLs - make the asset have a new image) and names/tickers (re-brands), which is why I ultimately decided to make the whole metadata mutable in the CATalog CHIP.
 * As new layers/kinds of CATs are developed, there would ideally be a way to add attributes and their value in the way CATs are mapped to details. For example, if Colored CATs become a thing, issuers should be able to set `color` to `blue`/`red`/etc. (so wallets supporting Colored CATs pick that up as well).
 * If someone wants to make the CAT data immutable, they can make an inner puzzle that locks some of those attributes (or send the title singleton to a burn address, if they also want to stop issuance/melting/revocation/etc.)

EvanWinget:
Thanks for the additional feedback, @Quexington 

Despite having some concerns and preferences on _convention_ for Titled CAT mutability: I've moved the CHIP to mutable-by-default. Immutability is opt-in by the issuer rather than imposed by the standard, and your broken-URI case is handled directly since a hash and its URI list can be updated together when the content genuinely changes.

On the discovery and the dusting concern: my understanding is that given a generic CAT, you can't tell from the coin whether it's titled because the asset ID is an opaque hash of the TAIL and the TAIL reveal only appears at mint and melt. The naive way to check is to register interest in a TAIL or hint hash and watch for the title singleton, which is a lot of subscriptions and, worse, a shared hash many wallets watch is a juicy dusting target. Please correct me if this is off.
  
The solutions I'm thinking about are:
1. Self-healing hint. An aware wallet attaches the launcher ID `L` as a memo when it sends a Unit, and when it receives a Unit with no `L` it resolves `L` once (via lineage walk) and re-attaches it on the next send, so the ecosystem converges to `L`-bearing coins on its own. `L` rides as a memo on an ordinary CAT spend, so there's no extra layer, no per-transfer cost, and a legacy wallet still spends the coin as a plain CAT.
2. A trustless index. A service maps asset ID to `L` by scanning `everything_with_singleton` mints, where the TAIL reveal is on-chain. Any wallet then answers "is this titled" with one fast lookup that re-verifies against the launcher binding, so the index is a cache rather than a registry you have to trust. It's the same shape of lookup a wallet already does to name any CAT.

Neither path subscribes to a shared hash, so neither creates the dusting target you flagged. A wallet reads `L` off its own coins or queries an index by asset ID.

The absence of a hint doesn't prove a CAT is untitled, so the cold case still costs a one-time lineage walk. I did consider a puzzle-enforced hint for a hard on-coin guarantee, but a layer below the TAIL would stop legacy wallets from spending the coin, which breaks the "looks like an ordinary CAT" property for a minimal title, so I'm inclined to keep it convention plus index and only consider enforcement for titles that are already aware-only such as revocable ones. 

If you have any feedback here I'm very open to suggestions.

EvanWinget:
Thanks for the review and feedback, @Yakuhito. I've updated the CHIP based on feedback from both you and Quex:

* Applicability: You're right. A Unit's asset ID is the tree hash of the `everything_with_singleton` TAIL curried to the title singleton's launcher, so the standard only applies to CATs minted under that TAIL. This is inherent to binding supply into the coin. I've added an Applicability section with this info and pointing to CATalog as the right tool for the existing CATs and all other TAILs.
* `precision` as a decimal count: Adopted. `0` = indivisible, `3` = current default, making power-of-ten structural and aligning with your `p` field in CATalog.
* Multiple precision options: Adopted your `pt`/`pn` idea (as `name_p`/`ticker_p`) for divisible titles.
* Royalty/revocability in the title: I kept them because you can't tell from the coin alone whether a Unit missing a layer is bare or stripped, since stripped and canonical Units share an asset ID.
* Immutability: You and @Quexington have a lot more experience than me, so I've shifted to the model you both suggested. Metadata is now mutable by default and the attributes that are fixed are: the Unit-to-title binding, the supply invariants, and the Unit layer configuration. I do hope that immutability becomes the expected convention though.
* Extensible attributes: this is good feedback - I need to think more about this.

Thank you - this was very useful!

EvanWinget:
Following up on solutions for other TAILs after a chat with Claude: I think the title binding generalizes to single-issuance (`genesis_by_coin_id`) for new mints, though I'd value folks poking holes before I update the CHIP.

Construction (using the standard launcher):
* Spend one setup coin `S` in a bundle that creates both the first CAT coin (parent `S`, so `asset_id = hash(genesis_by_coin_id(S))`) and a standard singleton launcher coin (giving the title singleton `L`). 
* The title is the singleton whose launcher was created in `S`'s spend. `genesis_by_coin_id` forbids any further mint or melt, meaning the singleton has zero supply authority. It is a pure metadata anchor that can update metadata or transfer title without ever touching the CAT, so the single-issuance fixed-supply guarantee is fully preserved.
* The spec would require exactly one launcher-puzzle-hash output in the genesis spend.

Binding and discovery:
 * The asset ID commits to `S`, and only `S`'s actual spend created the genuine launcher `L`. That spend is immutable, so the binding is unforgeable: an attacker's singleton can't prove `S`'s spend created it, and `S` is spent once so there's no second mint or collision.
* A wallet recovers a candidate `S` (from a memo hint or lineage walk), checks `hash(genesis_by_coin_id(S)) == asset_id` (self-verifying), then reads `S`'s spend to find `L`. This is one hop more than the `everything_with_singleton` case but still trustless and cacheable.

Yakuhito:
This special-casing reminds me of @Rigidity 's idea I mentioned in the CATalog CHIP call - in short, you could change the CATalog registration logic and have a custom approver puzzle per TAIL kind. So, for singleton-issued CATs, the singleton could just send a message to register the TAIL it controls, and, for single-issuance CATs, the puzzle of the coin that issued the CAT can be used to identify the issuer. This has the downside of needing someone to maintain and update the list of approver puzzles (which is my main reason for opting out of it), but I think it can get rid of registration fees.

CHIPs/chip-evanwinget-titled-cat.md
```diff
@@ -0,0 +1,379 @@
...
+
+### Technical feasibility
+
+Titled CAT is feasible today with no protocol changes. Everything the non-royalty core needs already exists and is deployed: CAT2, the CHIP-40 `everything_with_singleton` TAIL (Final, in production for Revocable CATs), the CHIP-38 revocation layer (Final), and the Singleton and Offer standards. The only new puzzle is the Title Singleton inner puzzle for which audited reference patterns exist. The optional royalty path additionally relies on the CHIP-56 Fee Layer, which is still an open proposal (PR #194). The composition occurs entirely at the CLVM / puzzle-hash level, so no consensus or VM change is required.
```
honyocokrokusuma-RLTM, line 100: Left #194

